PR #13, opened 2015-07-12
STM32F2 Support

rixth:
* Added STM32F2 support (all devices)
* Cleaned up README.md (organized it a bit more & fixed typos/grammar)

I tested `stm32-blinky`, `stm32-newlib` and `stm32-template` examples on a `STM32F207VF`. I was not able to test ChibiOS as I am not familiar with it.

There looks like a lot of changes, but some of it is whitespace. [View diff without whitespace](https://github.com/ObKo/stm32-cmake/pull/13/files?w=1)


ObKo:
Looks good, thank you for your contribution.

PR #15, opened 2015-10-11
Fix missing gcc_stm32.cmake include in CMakeLists templates.

alexisduque:
GCC STM32 Cmake toolchain file should be included in the CMakeLists project.


kylemanna:
Are the includes necessary?  I thought it would automatically include the toolchain file when the project was initialized?

alexisduque:
How do you initialize the project ? 
For instance, I run : 
 `cmake -DSTM32_CHIP=STM32F103VD -DCMAKE_TOOLCHAIN_FILE=../cmake/gcc_stm32.cmake -DSTM32Cube_DIR=/opt/STM32Cube_FW_F1_V1.2.0 -DCMAKE_MODULE_PATH=../cmake/ -DCMAKE_BUILD_TYPE=Debug ./ `
in `stm32-blinky/` folder and all are undefined.

kylemanna:
I've used this with sucess:

    mkdir build && cd build
    ❯ cmake -DSTM32_CHIP=STM32F103VG -DCMAKE_MODULE_PATH=$PWD/../../cmake -DCMAKE_TOOLCHAIN_FILE=$PWD/../../cmake/gcc_stm32.cmake ..
    -- No TOOLCHAIN_PREFIX specified, using default: /usr
    -- No TARGET_TRIPLET specified, using default: arm-none-eabi
    -- No STM32_FAMILY specified, trying to get it from STM32_CHIP
    -- Selected STM32 family: F1
    -- The ASM compiler identification is GNU
    -- Found assembler: /usr/bin/arm-none-eabi-gcc
    -- STM32F103VG is 103xG device
    -- No STM32Cube_DIR specified, using default: /opt/STM32Cube_FW_F1_V1.1.0
    -- Found CMSIS: /opt/STM32Cube_FW_F1_V1.1.0/Drivers/CMSIS/Device/ST/STM32F1xx/Include;/opt/STM32Cube_FW_F1_V1.1.0/Drivers/CMSIS/Include  
    -- Found STM32HAL: /opt/STM32Cube_FW_F1_V1.1.0/Drivers/STM32F1xx_HAL_Driver/Inc  
    -- No linker script specified, generating default
    -- STM32F103VG has 1024KiB of flash memory and 96KiB of RAM
    -- Configuring done
    -- Generating done

I think that perhaps you just need absolute path.

Have you tried your patch with `${CMAKE_CURRENT_SOURCE_DIR}/../cmake` or make `${CMAKE_SOURCE_DIR}/cmake` ?

alexisduque:
You're right, I need absolute path so this patch is unnecessary.

kylemanna:
I think defining the toolchain is handy if it's not set, I just think the `include` is unnecessary?

On some of my projects, if I see the `-DSTM_CHIP` variable I set the toolchain path if unset.

alexisduque:
Ok, so do you mean I should update my PR, to remove the `INCLUDE` and just add
```cmake
IF(NOT CMAKE_TOOLCHAIN_FILE)
  SET(CMAKE_TOOLCHAIN_FILE "gcc_stm32.cmake")
  MESSAGE(STATUS "No cmake toolchain file specified, using default: ${CMAKE_TOOLCHAIN_FILE}")
ENDIF()
```
?

kylemanna:
> Ok, so do you mean I should update my PR, to remove the INCLUDE and just add

That + With the full path to the CMAKE_TOOLCHAIN_FILE I think would be helpful.

alexisduque:
:smile:

alexisduque:
I removed the `INCLUDE`  and change to
```cmake
IF(NOT CMAKE_TOOLCHAIN_FILE)
  SET(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/../cmake/gcc_stm32.cmake")
  MESSAGE(STATUS "No cmake toolchain file specified, using default: ${CMAKE_TOOLCHAIN_FILE}")
ENDIF()
```
then run `cmake -DSTM32_CHIP=STM32F103VD -DSTM32Cube_DIR=/opt/STM32Cube_FW_F1_V1.2.0 -DCMAKE_MODULE_PATH="${PWD}/../../cmake/" -DCMAKE_BUILD_TYPE=Debug ..
`
And get `Unknown CMake command "STM32_GET_CHIP_TYPE` + `/usr/bin/cc` as compiler

Then I just add the  `INCLUDE` (as in my PR) and it works. Do you have any idea what's wrong ?

rpavlik:
By that time, it's most likely too late to specify the toolchain - the toolchain is really meant to be specified on the initial command line,and definitely before the project line (because otherwise the compilers, etc. are already set in stone)

FWIW, I actually use the presence/absence of the toolchain file in a CMake invocation to tell me whether I should run my "bootstrap" "superbuild" (that builds using the toolchain file with cmake's ExternalProject module, making an "easy" user-friendly configuration process) or proceed with the normal actual build.

ObKo:
Closing this because outdated.

PR #19, opened 2016-01-17
Add support F0 chips

JocusSoft:
Support of F0 family chips
Add print ram usage of output binary


ObKo:
Looks good. Merging it. It would be nice if you also port examples to some F0-based board...

JocusSoft:
OK. I write support F0 for templates

PR #22, opened 2016-02-02
Add support build project on MsWindows

novikovas:
Add CubeMX example


mgiaco:
I can confirm that this pull works - so please merge gcc_stm32.cmake it is the only file which is really 
required

ObKo:
Done in latest commits.

PR #23, opened 2016-02-18
Support for BSP

ma11:
The discovery boards of ST usually comes with several peripherals to play with. The drivers of these peripherals are shipped in STM32Cube* and it could ease development a lot to have cmake knowing these drivers.
I propose a pull which include the BSP drivers of the STM32F429I-DISCO board only, but of course it is then possible to add support for other boards.
The FindSTM32BSP.cmake file does the job and is used similarly to FindSTM32HAL.cmake.


ObKo:
Hello and sorry for the delay in replying!

I promoting support for BSP in stm32-cmake, but there is two issues with your patches:
* cmake/FindCMSIS.cmake - you'd hardcoded search path for CMSIS for F4 family. For what? Device-specific include dir is founded few rows above and stored in ${CMSIS_DEVICE_INCLUDE_DIR}
* cmake/gcc_stm32.cmake - Why had you added all those libraries to debug flags? They're project-specific and should be added in project's CMakeLists.txt if needed (using TARGET_LINK_LIBRARIES). Same for rdimon (semihosting).

Also, why had you added command for .elf file generation? GCC output is .elf files by default (just without extension on *nix).

ma11:
Hello ObKo, and thank you for your reply.

You are totally right about the issues you noticed. I'll correct them.
Thanks again for your review!

PR #31, opened 2016-09-06
add example  stm32-STM32F103RB-Nucleo-TIM

mgiaco:
Hello,

Today I add a basic example to show how this lib can be used with the original demo project source from the STM32Cube_FW_F1_xxx projects. So, therefore I copied the STM32F103RB-Nucleo TIM_TimeBase Inc and Src folder.

To start with this:
- Install python and pyinvoke (http://www.pyinvoke.org/ version >= 0.13)
- Install GNU ARM, and GNU ARM Tools from (http://gnuarmeclipse.github.io/)
- invoke configure
- invoke all
- invoke -l  to show the other targets

I will share it here maybe you will add it or ignore it. By the way with the new CMake 3.6.x CMAKE_FORE_COMPILER is deprecated so there is a patch needed in the future.


mgiaco:
Hi, 
Did you check this Pull? It would be really nice if you at least add the changes from cmake/gcc_stm32.cmake. Without this modification it is not possible to build it on windows.

ObKo:
Sorry for absense. Added windows executable support in de92ace183d3b85734ec88cc53c519c13992ef75 

Can you test it on Windows? 

Regarding to example - yes, I really don't want to merge it, because it's huge, unclear and has some build scripts...

mgiaco:
Thanks no problem.

PR #32, opened 2016-11-20
Fixed a missing ChibiOS file 

janoc:
Fixed a missing ChibiOS file causing link errors when using serial_usb module.


janoc:
Thanks!

PR #36, opened 2017-01-11
fix windows build

mgiaco:
Hello,

So today I tried your changes for windows but that does not work. So for me the variable was empty. I think if someone will use this variable it must be set before usage. So this fixes the build. Have you tried your changes on windows?

IF (WIN32)
set(CMAKE_EXECUTABLE_SUFFIX ".exe")
ENDIF()


mgiaco:
Sorry it was not possible for me delete the commit history. It was also not possible for me to cherry pick only the cmake/gcc_stm32.cmake file for this pull.

cheers

ObKo:
Done in latest commit

PR #55, opened 2017-06-12
Moved LED to TIM1/PA10 on F1 because TIM8 not supported an all devices

clausgf:
Low/Medium density STM32F1 devices like my STM32F103CBT6 don't support TIM8, therefore I moved the LED output to TIM1/PA10.


ObKo:
Closing this because outdated, stm32-cmake rework in progress: https://github.com/ObKo/stm32-cmake/pull/117, blinky example will use SysTick there.

Issue #82, opened 2018-08-23
Is modern cmake would be possible in embedded world ?


davidtazy:
Hi,
(I am a "desktop c++ dev" )

cmake >3.0 invite us to see packages  as "TARGET"

1. **would it be possible to have CMSIS, HAL... as target and be able to write:**
![capture](https://user-images.githubusercontent.com/30446373/44556806-bb2d0900-a73b-11e8-9f69-fe311469f1cb.PNG)

```
...


add_library(MySensorInterfaceLib ${files}
target_link_libraries(MySensorInterfaceLib  PRIVATE Stm32::CMSIS)

add_library(MyBusinessValueLib ${files}
target_link_libraries(MyBusinessValueLib PUBLIC MySensorInterfaceLib  )

add_executable(MyProject main.c)
target_link_libraries(MyProject PRIVATE MyBusinessValueLib ) 
```
i think it would be more "readable" and it can help low coupling

2. **as a general question, why embedded world encourage us to **not** create libraries but build all files ?**

excuse my english...

free-easy:
1. It's possible to use the target based approach. See my pull request.
It's quite trivial, I just define some IMPORTED target at the end of the Find module, where I gather all the variables.

2. I don't think there's any reason for the embedded world to encourage this.
Actually IMHO the embedded world is quite weird. 
Each manufacturer build their custom Eclipse-based IDE, the library they supply is often only accessible through their website, when now days it would be much useful to have those libraries (which are given with the source code) on platforms like GitHub, with another benefit that issue tracking is much easier compared to their ugly forum.
Therefore, I think the problem is more that the embedded world (in many cases) just ignores tools like CMake.

ObKo:
If you're still interested, please take a look at modern-cmake branch: https://github.com/ObKo/stm32-cmake/tree/modern-cmake

I've implemented support for CMSIS and HAL using imported targets and modern cmake features, such as target_sources and INTERFACE scope.

Looks pretty good for me, but I need more opinions and suggestions.

Will make that master after implementing some missing features.

davidtazy:
Still interested!
Thx you for that.  I will look at it!

PR #87, opened 2018-12-12
feat(STM32Cube): supprot modern CMake's target-based paradigm

chenlijun99:
Related to #82


mgiaco:
Hi, Do you have a sample project where we can see the benefit of this pull.
thanks

free-easy:
Hi,
here I show you a quick and dirty snippet that should clarify the benefit of the target based approaach.

Before (example taken from stm32-blinky)
```cmake
PROJECT(stm32-blinky)

CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
ENABLE_LANGUAGE(ASM)

FIND_PACKAGE(CMSIS REQUIRED)
FIND_PACKAGE(STM32HAL COMPONENTS gpio tim REQUIRED)

INCLUDE_DIRECTORIES(
    ${CMAKE_CURRENT_SOURCE_DIR}
    ${CMSIS_INCLUDE_DIRS}
    ${STM32HAL_INCLUDE_DIR}
)

SET(PROJECT_SOURCES
    main.c
)

ADD_EXECUTABLE(${CMAKE_PROJECT_NAME} ${PROJECT_SOURCES} ${CMSIS_SOURCES} ${STM32HAL_SOURCES})

STM32_SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME})
STM32_ADD_HEX_BIN_TARGETS(${CMAKE_PROJECT_NAME})
```
After

```cmake
PROJECT(stm32-blinky)

CMAKE_MINIMUM_REQUIRED(VERSION 3.0)
ENABLE_LANGUAGE(ASM)

# Notice here that I don't have to invoke FIND_PACKEGE(CMSIS) anymore, because it's 
# up to STM32HAL's find module to know that STM32HAL depends on CMSIS.
# Each dependency should bring in its own dependencies automatically IMHO
# And the target based approch simplifies immensely the way to achieve this.
FIND_PACKAGE(STM32HAL COMPONENTS gpio tim REQUIRED)

SET(PROJECT_SOURCES
    main.c
)

ADD_EXECUTABLE(${CMAKE_PROJECT_NAME} ${PROJECT_SOURCES})

# Sidenote: Modern CMake encourages to bind build parameters to targets specifically, 
# instead of polluting the global "parameter space"
TARGET_INCLUDE_DIRECTORIES(${CMAKE_PROJECT_NAME}
    PRIVATE
    ${CMAKE_CURRENT_SOURCE_DIR}
)

# **Here is the key point of the pull request**
# It's simply enough to say that my executable (or more generally, this target) links
# to a target called STM32Cube::HAL and all the include directories, sources declared
# PUBLIC or INTERFACE by the target are bound to this target.
# Furthermore, if STM32Cube::HAL links in a PUBLIC way to other library targets,
# the build parameters declared PUBLIC, INTERFACE by them will be transitively
# added.
# Therefore, since STM32Cube::HAL links to STM32Cube::CMSIS publicly,
# this executable target will get to know the build parameters required by STM32Cube::CMSIS
# too, without the developer to explicitly write anything.
TARGET_LINK_LIBRARIES(${CMAKE_PROJECT_NAME} STM32Cube::HAL)

STM32_SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME})
STM32_ADD_HEX_BIN_TARGETS(${CMAKE_PROJECT_NAME})
```

free-easy:
I admit that there are a couple of things that can be perfected with this pull request, such as updating the README.md, adding a sample project, supporting the BSP (and even the other Find modules too), etc.

If you tell me that you're in favor of this target based approach, I'll try to get some time during the Christmas holidays to work further on this.

Obviously, if you see errors in my explanation of the target based workflow, please correct me.
CMake's official documentation is very limited in terms of how to use it. It often just describes something, without giving an advise on how that thing can be effectively used.

mgiaco:
@free-easy  many thanks looks interesting.  But how dose this work with STM32Cube::HAL. Do you have any good link with some examples in general. I am wondering how this link works. How dose cmake knows where STM32Cube::CMSIS could be found.

free-easy:
@mgiaco Here is an [article](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/
) that I read months ago and found quite useful.
If you want a deeper insight of the topic there always the official documentation ([find_package](https://cmake.org/cmake/help/v3.8/command/find_package.html?highlight=i), [target](https://cmake.org/cmake/help/v3.10/manual/cmake-buildsystem.7.html#introduction)). I admin that I've never read them thoroughly.

Anyway, let me give you (at least I try) an explanation for this specific case.
`STM32Cube::HAL` is just a name. Generally it's good practise to given namespaced target names and `STM32Cube` is in this case the namespace.

This is the core part of this pull request

FindSTM32HAL.cmake
```cmake
# In the FindSTM32HAL, before finding STM32HAL, I tell to find CMSIS first, because, as Find module 
# provider, I and only I have to know that STM32HAL depends on CMSIS.
find(CMSIS REQUIRED)
if (CMSIS_FOUND)
    # Execute the good old FindSTM32HAL.cmake
    ...
endif() 

if(STM32HAL_FOUND AND NOT TARGET STM32Cube::HAL)
    # If STM32HAL has been found, I just define a target called STM32Cube::HAL
    add_library(STM32Cube::HAL INTERFACE IMPORTED)
    # and I say declare the build parameters for this target
    # Notice here that I've added STM32Cube::CMSIS as INTERFACE_LINK_LIBRARIES
    # thsi tell CMake that when a target links to STM32Cube::HAL, it has to link to STM32Cube::CMSIS
    # too.
    set_target_properties(STM32Cube::HAL 
        PROPERTIES
        INTERFACE_LINK_LIBRARIES STM32Cube::CMSIS
        INTERFACE_INCLUDE_DIRECTORIES "${STM32HAL_INCLUDE_DIRS}"
        INTERFACE_SOURCES "${STM32HAL_SOURCES}"
        )
endif()
```

Therefore, CMake doesn't need to know anything magically. It everything explicitly written by the Find module provider, which is this repository.

Hish15:
This is the way to go with CMake there is no denying that!
Good work  @free-easy!

cmake/FindCMSIS.cmake
```diff
@@ -156,8 +156,20 @@ IF(STM32_CHIP_TYPE)
         CMAKE_FIND_ROOT_PATH_BOTH
     )
     LIST(APPEND CMSIS_SOURCES ${CMSIS_STARTUP_SOURCE_FILE})
+
+    SET(CMSIS_COMPILE_DEFINITIONS "STM32F${STM32_CHIP_TYPE}")
```
Hish15: Is it normal to have a F here? STM32**F**${STM32_CHIP_TYPE}

cmake/FindSTM32HAL.cmake
```diff
@@ -1,176 +1,190 @@
...
+        SET(HAL_PREFIX stm32l0xx_)
+    ELSEIF(STM32_FAMILY STREQUAL "L1")
+        SET(HAL_COMPONENTS adc comp cortex crc cryp dac dma flash flash_ramfunc
+    					   gpio i2c i2s irda iwdg lcd nor opamp pcd pwr rcc rtc
+    					   sd smartcard spi sram tim uart usart wwdg)
+        SET(HAL_REQUIRED_COMPONENTS cortex pwr)
+        
+        # Components that have _ex sources
+        SET(HAL_EX_COMPONENTS adc cryp dac flash opamp pcd pwr rcc rtc spi tim)
+        # Components that have ll_ in names instead of hal_
```
Hish15: Is the list missing or the comment useless?

cmake/FindCMSIS.cmake
```diff
@@ -156,8 +156,20 @@ IF(STM32_CHIP_TYPE)
...
 ENDIF()
 
 INCLUDE(FindPackageHandleStandardArgs)
 
 FIND_PACKAGE_HANDLE_STANDARD_ARGS(CMSIS DEFAULT_MSG CMSIS_INCLUDE_DIRS CMSIS_SOURCES)
+
+IF(CMSIS_FOUND AND NOT TARGET STM32Cube::CMSIS)
+    ADD_LIBRARY(STM32Cube::CMSIS INTERFACE IMPORTED)
+    SET_TARGET_PROPERTIES(STM32Cube::CMSIS
```
Hish15: Why Don't you use ?
```
target_include_directories()
target_sources()
target_compile_definitions()
```

cmake/FindSTM32HAL.cmake
```diff
@@ -1,176 +1,190 @@
...
+    ENDFOREACH()
+ENDIF()
 
 INCLUDE(FindPackageHandleStandardArgs)
 
 FIND_PACKAGE_HANDLE_STANDARD_ARGS(STM32HAL DEFAULT_MSG STM32HAL_INCLUDE_DIR STM32HAL_SOURCES)
+
+if(STM32HAL_FOUND AND NOT TARGET STM32Cube::HAL)
+    add_library(STM32Cube::HAL INTERFACE IMPORTED)
+    set_target_properties(STM32Cube::HAL
```
Hish15: better use the target_blabla() functions to set target properties

cmake/FindCMSIS.cmake
free-easy: It's an error. I just recently caught this bug when in one of the projects I started using L0 family devices.

cmake/FindSTM32HAL.cmake
free-easy: I don't think it's related to my PR. This comment is present in the current master too.

cmake/FindCMSIS.cmake
free-easy: TL;DR; You're right. 

At the time of this PR, I just followed kind of blindly the article that I mentioned in https://github.com/ObKo/stm32-cmake/pull/87#issuecomment-448932193.

The fact is that when one declares `IMPORTED` targets, functions like `target_*()` can't be used and we have to modify the target using the low-level `set_target_properties()` function.
But actually in this case AFAIK, since the target that we're declaring is INTERFACE, using `IMPORTED` or not doesn't matter much. `IMPORTED` targets should really be used when we have a library that is already built somewhere.

As of now, I think we should remove the `IMPORTED` specifier and declare the targets as `STATIC`, for the following reason: when declaring some target as INTERFACE we're just saying that everything that links to that target has to take its properties too; specifically in our case, where we have INTERFACE targets that include a series of sources, all the targets that link to these INTERFACE targets will compile those source files. While this might be useful when we deal with targets like STM32HAL (which requires a `stm32fxhal_conf.h` to lay somewhere) and generally with targets that are not self-contained, since everything the INTERFACE target specifies is compiled together with the target that links it, allowing us to define the "missing pieces" of the INTERFACE target in the target that links it, the INTERFACE target stuff breaks really quickly once the build system starts getting a little complex.

Imagine when we want to link the STM32HAL target to multiple executables. Or if we have our project organized into modules and each module should be as self-contained as possible, i.e. should specify all its dependencies. Using the INTERFACE target method, for every target that links the INTERFACE target, the sources files specified for the INTERFACE target are re-compiled, which is really bad.
Using the STATIC target method, we have instead that the target is built stand-alone and the produced *.a library file is linked to every target that links it.
The only drawback of STATIC targets is that if they're not really stand-alone, well... they can't be built stand-alone... Which is indeed the case for STM32HAL, which requires the presence of `stm32xx_hal_conf.h` to complete the puzzle.

Currently, in order to solve this problem, I require the user to define a target (e.g. for the case of STM32HAL, let's say that I require a target called `STM32HALConf`, which must specify the include directory where `stm32xx_hal_conf.h` can be found) before calling `find_package`, and then the target that I define in the find module will link the target provided by the user.

Let's just say that, while the INTERFACE target method works by telling the user "I give the incomplete puzzle, do as you want in the application layer", the STATIC target method works by telling the user "Give me this missing piece and I'll give you the nice, self-contained target that you just need to blindly link".

cmake/FindSTM32HAL.cmake
free-easy: See https://github.com/ObKo/stm32-cmake/pull/87#discussion_r322024699

ObKo:
If you're still interested, please take a look at modern-cmake branch: https://github.com/ObKo/stm32-cmake/tree/modern-cmake

I've implemented support for CMSIS and HAL using imported targets and modern cmake features, such as target_sources and INTERFACE scope.

Looks pretty good for me, but I need more opinions and suggestions.

Will make that master after implementing some missing features.

Hish15:
@ObKo This is great. 
I'm willing to check your changes to use modern-cmake,
can you open a PR so we can check and comment?

ObKo:
https://github.com/ObKo/stm32-cmake/pull/117

PR #117, opened 2020-05-17
Modern CMake paradigm support

ObKo:
stm32-cmake was started 8 years ago as a pet project for making bare-metal development more easily.

I did not spend much time for this project past years. Surprisingly, it became quite popular.

Things changed, especially, cmake changed greatly, and tricks used in current stm32-cmake scripts are considered ugly. There was some discussion about that: https://github.com/ObKo/stm32-cmake/issues/82 https://github.com/ObKo/stm32-cmake/pull/87

So, past several month I thought carefully and re-implemented whole project using modern cmake paradigm (targets), which also fixed some inborn design issues - one chip/family per project, incomplete family support, etc.

With new paradigm typical cmake project will look like:
```cmake
cmake_minimum_required(VERSION 3.8)

project(stm32-template C ASM)

find_package(CMSIS COMPONENTS STM32F407VG REQUIRED)

add_executable(stm32-template.elf main.c)
target_link_libraries(stm32-template.elf CMSIS::STM32::F407VG STM32::NoSys)
```
All features are formed as imported targets (libraries). When you link your executable with CMSIS::STM32::F407VG cmake will add startup sources, linker script to your executable. HAL can be linked same way.

For now I have added support for CMSIS and HAL for almost all STM32 families: F0, G0, L0, F1, L1, F2, F3, F4, G4, L4, F7, H7.  Other features are missing for now, some places should be reworked (linker script generation, for example). After that I will merge this into master.

I collect opinions and suggestions about what features to implement and what to change.


cmake/FindCMSIS.cmake
```diff
@@ -1,196 +1,164 @@
...
-    STRING(MAKE_C_IDENTIFIER "${SRC}" SRC_CLEAN)
-    SET(CMSIS_${SRC_CLEAN}_FILE SRC_FILE-NOTFOUND)
-    FIND_FILE(CMSIS_${SRC_CLEAN}_FILE ${SRC}
-        PATH_SUFFIXES src stm32${STM32_FAMILY_LOWER} cmsis
-        HINTS ${STM32Cube_DIR}/Drivers/CMSIS/Device/ST/STM32${STM32_FAMILY}xx/Source/Templates/
-        CMAKE_FIND_ROOT_PATH_BOTH
+if(NOT CMSIS_FIND_COMPONENTS)
+    set(CMSIS_FIND_COMPONENTS 
+        STM32F0 STM32G0 STM32L0 STM32F1 STM32L1 STM32F2 STM32F3 STM32F4 STM32G4 
+        STM32L4 STM32F7 STM32H7
```
Hish15, line 4: For readability I would do : 
```suggestion
        STM32F0 STM32F1 STM32F2 STM32F3 STM32F4 STM32F7
        STM32G0 STM32G4        
        STM32H7
        STM32L0 STM32L1 STM32L4         
```

cmake/FindHAL.cmake
```diff
@@ -0,0 +1,298 @@
...
+    sd sdram smartcard smbus spdifrx spi sram swpmi tim uart usart wwdg
+)
+set(HAL_EX_DRIVERS_H7
+    adc crc cryp dac dfsdm dma eth flash hash i2c i2s ltdc mmc opamp pcd pwr rcc 
+    rng rtc sai sd smartcard spi tim uart usart
+)
+set(HAL_LL_DRIVERS_H7
+    adc bdma comp crc crs dac delayblock dma dma2d exti fmc gpio hrtim i2c lptim 
+    lpuart mdma opamp pwr rcc rng rtc sdmmc spi swpmi tim usart usb utils
+)
```
Hish15, line 158: The most logic order here is family/index: F0...F7, G0..,

cmake/FindHAL.cmake
```diff
@@ -0,0 +1,298 @@
...
+
+if(NOT HAL_FIND_COMPONENTS)
+    set(HAL_FIND_COMPONENTS 
+        STM32F0 STM32G0 STM32L0 STM32F1 STM32L1 STM32F2 STM32F3 STM32F4 STM32G4 
+        STM32L4 STM32F7 STM32H7
```
Hish15, line 163: order by family then index instead

cmake/FindHAL.cmake
```diff
@@ -0,0 +1,298 @@
...
+    
+    if ((NOT HAL_${FAMILY}_INCLUDE) OR (NOT HAL_${FAMILY}_SOURCE))
+        set(HAL_${COMP}_FOUND FALSE)
+        continue()
+    endif()
+
+    if(NOT (TARGET HAL::STM32::${FAMILY}))
+        add_library(HAL::STM32::${FAMILY} INTERFACE IMPORTED)
+        target_link_libraries(HAL::STM32::${FAMILY} INTERFACE STM32::${FAMILY})
+        target_link_libraries(HAL::STM32::${FAMILY} INTERFACE CMSIS::STM32::${FAMILY})
```
Hish15, line 213:
```suggestion
        target_link_libraries(HAL::STM32::${FAMILY} INTERFACE   STM32::${FAMILY}
                                                                CMSIS::STM32::${FAMILY})
```

cmake/FindHAL.cmake
```diff
@@ -0,0 +1,298 @@
...
+        string(TOUPPER ${DRV} DRV)
+        
+        find_file(HAL_${FAMILY}_${DRV}_LL_SOURCE
+            NAMES stm32${FAMILY_L}xx_ll_${DRV_L}.c
+            PATHS "${HAL_${FAMILY}_PATH}/Src"
+            NO_DEFAULT_PATH
+        )
+        list(APPEND HAL_${FAMILY}_SOURCES "${HAL_${FAMILY}_${DRV}_LL_SOURCE}")
+        if(NOT HAL_${FAMILY}_${DRV}_LL_SOURCE)
+            message(WARNING "Cannot found LL_${DRV} driver for ${COMP}")
```
Hish15, line 268:
```suggestion
            message(WARNING "Cannot find LL_${DRV} driver for ${COMP}")
```

cmake/stm32/common.cmake
```diff
@@ -0,0 +1,162 @@
+set(STM32_SUPPORTED_FAMILIES L0 L1 L4 F0 F1 F2 F3 F4 G4 L4 F7 H7)
```
Hish15, line 1: G0 is supported too :)
```suggestion
set(STM32_SUPPORTED_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4)

```

cmake/stm32/common.cmake
```diff
@@ -0,0 +1,162 @@
...
+endfunction()
+
+function(stm32_get_chip_info CHIP FAMILY TYPE DEVICE)
+    string(TOUPPER ${CHIP} CHIP)
+        
+    string(REGEX MATCH "^STM32([A-Z][0-9])([0-9A-Z][0-9][A-Z][0-9A-Z]).*$" CHIP ${CHIP})
+    
+    if((NOT CMAKE_MATCH_1) OR (NOT CMAKE_MATCH_2))
+        message(FATAL_ERROR "Unknown chip ${CHIP}")
+    endif()
```
Hish15, line 51: Making this a function seems mandatory. 
This is performed multiple times in multiple places so updating will be trickier

cmake/stm32/common.cmake
```diff
@@ -0,0 +1,162 @@
...
+        set(FLASH "1024K")
+    elseif(SIZE_CODE STREQUAL "H")
+        set(FLASH "1536K")
+    elseif(SIZE_CODE STREQUAL "I")
+        set(FLASH "2048K")
+    elseif(SIZE_CODE STREQUAL "Z")
+        set(FLASH "192K")
+    else()
+        set(FLASH "16K")
+        message(WARNING "Unknow flash size for device ${DEVICE}")
```
Hish15, line 102:
```suggestion
        message(WARNING "Unknow flash size for device ${DEVICE}". Set to ${FLASH}.)

```

cmake/stm32/common.cmake
```diff
@@ -0,0 +1,162 @@
...
+    
+    if(FAMILY STREQUAL "F1")
+        stm32f1_get_memory_info(${DEVICE} ${TYPE} FLASH RAM)
+    elseif(FAMILY STREQUAL "L1")
+        stm32l1_get_memory_info(${DEVICE} ${TYPE} FLASH RAM)
+    elseif(FAMILY STREQUAL "F2")
+        stm32f2_get_memory_info(${DEVICE} ${TYPE} FLASH RAM)
+    elseif(FAMILY STREQUAL "F3")
+        stm32f3_get_memory_info(${DEVICE} ${TYPE} FLASH RAM)
+    endif()
```
Hish15, line 118: 1) Reorder.
2) Why only those families ?

cmake/stm32/common.cmake
```diff
@@ -0,0 +1,162 @@
...
+include(stm32/l0)
+include(stm32/f1)
+include(stm32/l1)
+include(stm32/f2)
+include(stm32/f3)
+include(stm32/f4)
+include(stm32/g4)
+include(stm32/l4)
+include(stm32/f7)
+include(stm32/h7)
```
Hish15, line 159:
```suggestion
include(stm32/f0)
include(stm32/f1)
include(stm32/f2)
include(stm32/f3)
include(stm32/f4)
include(stm32/f7)
include(stm32/g0)
include(stm32/g4)
include(stm32/h7)
include(stm32/l0)
include(stm32/l1)
include(stm32/l4)
```

cmake/stm32/f0.cmake
```diff
@@ -0,0 +1,25 @@
...
+)
+
+stm32_util_create_family_targets(F0)
+
+target_compile_options(STM32::F0 INTERFACE 
+    -mcpu=cortex-m0
+)
+target_link_options(STM32::F0 INTERFACE 
+    -mcpu=cortex-m0
+)
```
Hish15, line 25: Do we need both function?

examples/blinky/blinky.c
```diff
@@ -0,0 +1,58 @@
...
+    GPIO_InitTypeDef GPIO_Config;
+
+	GPIO_Config.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_Config.Pull = GPIO_NOPULL;
+	GPIO_Config.Speed = GPIO_SPEED_FREQ_HIGH;
+
+	GPIO_Config.Pin = LED_PIN;
+
+	LED_PORT_CLK_ENABLE();
+	HAL_GPIO_Init(LED_PORT, &GPIO_Config);
```
Hish15, line 44:
```suggestion
    GPIO_InitTypeDef GPIO_Config;
    
    GPIO_Config.Mode = GPIO_MODE_OUTPUT_PP;
    GPIO_Config.Pull = GPIO_NOPULL;
    GPIO_Config.Speed = GPIO_SPEED_FREQ_HIGH;
    
    GPIO_Config.Pin = LED_PIN;
    
    LED_PORT_CLK_ENABLE();
    HAL_GPIO_Init(LED_PORT, &GPIO_Config);
```

examples/blinky/blinky.c
```diff
@@ -0,0 +1,58 @@
...
+
+int main(void)
+{
+	HAL_Init();
+	initGPIO();
+	// 1kHz ticks
+	HAL_SYSTICK_Config(SystemCoreClock / 1000);
+
+	for (;;)
+		__WFI();
```
Hish15, line 55:
```suggestion
        __WFI();
```

tests/cmsis/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+cmake_minimum_required(VERSION 3.8)
+set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
+
+if(NOT TEST_FAMILIES)
+    set(TEST_FAMILIES F0 G0 L0 F1 L1 F2 F3 F4 G4 L4 F7 H7)
```
Hish15, line 5:
```suggestion
    set(TEST_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4)
```

tests/hal/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+cmake_minimum_required(VERSION 3.8)
+set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
+
+if(NOT TEST_FAMILIES)
+    set(TEST_FAMILIES F0 G0 L0 F1 L1 F2 F3 F4 G4 L4 F7 H7)
```
Hish15, line 5:
```suggestion
    set(TEST_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4)
```

MBetters:
It looks like `TOOLCHAIN_BIN_PATH` is used instead of `TOOLCHAIN_PREFIX` in `stm32_gcc.cmake`, so `README.md` needs to be updated to reflect that.

Hish15:
I noticed that BSP are no longer handled is that wanted?
see #118

cmake/stm32/common.cmake
ObKo: Those families have different RAM size inside one device type, so additional function needed to determine RAM size

ObKo:
I'm going to return BSP package, just need more time. Will think what to do with generated BSP.

cmake/FindBSP.cmake
```diff
@@ -0,0 +1,320 @@
...
+endif()
+
+foreach(COMP ${BSP_FIND_COMPONENTS})
+    string(TOLOWER ${COMP} COMP_L)
+    string(TOUPPER ${COMP} COMP_U)
+    
+    string(REGEX MATCH "^STM32([FGHL][0-9]).*$" COMP_U ${COMP_U})
+    
+    if(NOT CMAKE_MATCH_1)
+        message(FATAL_ERROR "Unknown HAL component: ${COMP}")
```
Hish15, line 237:
```suggestion
        message(FATAL_ERROR "Unknown BSP component: ${COMP}")
```

barafael:
Hey, thanks for doing this, it is working great for me so far. Are you planning to add support for the ST LL library, too?

Hish15:
Hi @barafael, actually LL libraries are already handled... what MCU are using?

barafael:
@Hish15 Never mind, I was doing it wrong.

I think once the modern-cmake support merges, it would be great to have some more example projects and documentation on how this works. As soon as I understand this better, I will try to do that.

Hish15:
@Obko, @barafael is right we do need to add more examples to make the whole thing easier to use, but also more appealing.
People are making PR to the futur old master branch, we better switch to the modern paradigm, before adding more feature :)

barafael:
I am a fan of #8d80b "STM32Cube packages as git modules."

However, this bloats the size of this repo quite a bit. For example, my continuous integration builds failed because the docker containers were too small.

Would it be possible to load just the required cube packages?

Hish15:
@barafael, a solution is not to recurse init when cloning or pulling in your CI build.

Another solution, is not to use submodule but Cmake to retireve the needed HAL from github directly.
[Cmake > 3.11](https://cmake.org/cmake/help/v3.14/module/FetchContent.html).

This is pretty neat since it allows to auto download if a HAL is not provided for instance.

barafael:
@Hish15 > not to recurse init: I had set up my CMakeLists.txt so that it uses the stm32-cmake submodule as STM32_CUBE_XY_PATH. Previously, my project had its own submodule for that. This means that I would need to clone recursive everytime i want to build. Perhaps a solution would be to only check out the desired submodule.

I think if somebody had many stm32 projects, and each has their own copy of ALL cube libraries for all families, it would eat a lot of disk space.

Hish15:
@barafael, you are totaly right.

the FetchContent method approach  seems to be a better solution here.

barafael:
related to #145 "ST lib´s as submodules"

mgiaco:
> I am a fan of #8d80b "STM32Cube packages as git modules."
> 
> However, this bloats the size of this repo quite a bit. For example, my continuous integration builds failed because the docker containers were too small.
> 
> Would it be possible to load just the required cube packages?



> @barafael, you are totally right.
> 
> the FetchContent method approach seems to be a better solution here.

But it would be great if someone can disable this. So fetch as an option would be great.

Hish15:
@mgiaco, The fetch would be mainly for the case of CI tests.
which can be also handled using docker... so many options haha.

barafael:
Should the build system even care about Cube etc.? For me, just pointing at the CUBE_XYZ directory from CMakeLists.txt is completely sufficient and leaves it open to the user how to fetch the Cube repositories.

ObKo:
So, I've added stm32_fetch_cube(<FAMILIES>) function that will fetch STM32Cube sources from ST's github. It will do nothing if STM32_CUBE_<FAMILY>_PATH variable is already exist.

barafael:
Just a suggestion: I am converting older projects which originally were generated using CubeMX. They have a top-level directory 'Drivers', which is the same as STM_CUBE_FAMILY_PATH/Drivers. But I think when I set STM_CUBE_FAMILY_PATH to '.' (the directory containing the directory 'Drivers', it does not work. I am working around by adding a submodule for Cube or FetchContent-ing it, but I think it would be nice if the generated source could be used by stm32-cmake.

ObKo:
> But setting STM_CUBE_FAMILY_PATH to '.' (the directory containing the directory 'Drivers'), it does not work.

Try to set it to `${CMAKE_CURRENT_SOURCE_DIR}` instead of '.', or use canonical path.

jhalmen:
i'm excited to see so much progress on this!

regarding the automatic fetching of the cube sources: these repositories are huge, and contain everything including documentation and loads of unnecessary (when only interested in building own code) examples.
wouldn't it maybe be possible to instead directly fetch only the necessary things?
what i have in mind is for example:
instead of
* https://github.com/STMicroelectronics/STM32CubeF7 repo (~450MB)

just download 
* the cmsis definitions (https://github.com/STMicroelectronics/cmsis_device_f7  <1MB)
* and the HAL/LL libraries (https://github.com/STMicroelectronics/stm32f7xx_hal_driver   <2MB)

this would speed up the time it would take to set up a new project, and would also save a lot of resources.

i may be mistaken, but i don't think one would need more than those 2 repos for each family.
would be awesome if you could maybe take a look at this. thanks!

ObKo:
> 
> 
> i'm excited to see so much progress on this!
> 
> regarding the automatic fetching of the cube sources: these repositories are huge, and contain everything including documentation and loads of unnecessary (when only interested in building own code) examples.
> wouldn't it maybe be possible to instead directly fetch only the necessary things?
> what i have in mind is for example:
> instead of
> 
>     * https://github.com/STMicroelectronics/STM32CubeF7 repo (~450MB)
> 
> 
> just download
> 
>     * the cmsis definitions (https://github.com/STMicroelectronics/cmsis_device_f7  <1MB)
> 
>     * and the HAL/LL libraries (https://github.com/STMicroelectronics/stm32f7xx_hal_driver   <2MB)
> 
> 
> this would speed up the time it would take to set up a new project, and would also save a lot of resources.
> 
> i may be mistaken, but i don't think one would need more than those 2 repos for each family.
> would be awesome if you could maybe take a look at this. thanks!

Take a look at d9cb37c70473f4387c71a2d965d12bb970b25724 now CMSIS and HAL can be used without Cube as separated packages. And there is FetchContent for CMSIS and HAL without whole Cube.

It takes ~650MB for all CMSISes & HALs.

barafael:
> Try to set it to `${CMAKE_CURRENT_SOURCE_DIR}` instead of '.', or use canonical path.

I could not get that working. Still says
Could NOT find CMSIS (missing: CMSIS_SOURCES STM32F0) (found version
  "0.0.0")
Should I open separate issue? Quite possible I am doing it wrong.

ObKo:
Finally merged this.

Now it known as stm32-cmake v2.x, old code goes to v1 branch.
Some features (ChibiOS, uGFX) are missing for now, will add it later.

geokon-gh:
This project is fantastic! Thank you so much for the hard work. You managed to make what I had been shooting for a couple years ago before I ran out of steam https://geokon-gh.github.io/bluepill/

My only suggestion at the moment is that the downloading of the CMSIS source into the build directory is very handy but a bit not "modern CMake". Each build directory ends up with another copy which is a bit clunky.

One alternative I'd just throw out there is to use something like Hunter that keeps a global dependency cache. https://github.com/ruslo/hunter . It might be a bit over-engineered for embedded development- but it's quite handy and it speeds up builds quite a bit. It lets you play around with build parameters much more easily and it also generates the namespacing for dependencies automatically so, at least partly, you don't need to create those manually (ex: CMSIS::STM32::F407VG)

I think @ruslo is also from Novosibirsk - maybe you could ping him and see if he'd be interested in helping out. His CGold tutorial is just great for groking CMake: https://cgold.readthedocs.io/ though based on your style I feel that maybe you've seen it before :)

Thanks again for this!

PR #124, opened 2020-06-19
Sets FreeRTOS heap source to NOTFOUND before calling find_file to fix…

alxhoff:
… find_file finding a false heap implementation


Hish15:
What is the point? This is already the behavior.
cmake's find_file doc states that : 
 If nothing is found, the result will be <VAR>-NOTFOUND

alxhoff:
On my system if I don't manually do this CMake finds heap_4.h for some reason.

ObKo:
> On my system if I don't manually do this CMake finds heap_4.h regardless of the implementation I specified for some strange reason.

Thats because find_file saves result into CMake cache, so find_file will be ignored during next run. So when changing heap implementation you will need to flush your CMake cache.

Not sure that there is the right way to fix that...

alxhoff:
ahhh. Maybe a manual unsetting of the cache variable?

Hish15:
Building on @ObKo answer, some vars are cached some are not.
If you change some cached var, you are excpected to clean cmake cache and rerun.

@alxhoff, Do you confirm that it is just a caching issue?

alxhoff:
Yes this is a caching issue. For my target applications (more so for my students) who are experimenting with different heap implementations and thus changing this one specific cached variable a lot it is easier to reset the one variable rather than clear the cache. While unsetting variables is not common practice it makes use easier for my applications. The question is that is this behavior better for the average user?

PR #127, opened 2020-06-22
Added fsmc and sdmmc to LL components for F4

alxhoff:
Also removed i2s from LL as it is not in LL, only HAL, at least for the F4.


alxhoff:
Included in #129

PR #130, opened 2020-06-22
TensorFlow Lite Micro cmake module

alxhoff:
Tested working in numerous STM32 projects.


ObKo:
Can you later re-implement this into new stm32-cmake using modern CMake features: https://github.com/ObKo/stm32-cmake/pull/117 ?

PR #132, opened 2020-06-23
Cmake module to find STM32 Discovery library files

alxhoff:


Hish15:
I think that I'm missing something, but what is the difference with whats its done in https://github.com/ObKo/stm32-cmake/blob/master/cmake/FindSTM32BSP.cmake ?

alxhoff:
I am not sure why STM does it but there is [this](https://my.st.com/content/my_st_com/en/products/embedded-software/mcu-mpu-embedded-software/stm32-embedded-software/stm32-standard-peripheral-library-expansion/stsw-stm32138.license=1593419758128.product=STSW-STM32138.version=1.0.1.html) firmware package for the discovery that shares some similarities with the BSP sources but then then for instance the l3gd20 sources are no longer discovery specific. You can probably achieve the same thing using the BSP sources but there is a specific discovery firmware package that this module is targeted towards.

alxhoff:
A good example is the `ioe` sources that don't actually exist in the BSP sources. They can be found in varying forms in the STD Peripherals library but then it becomes messy pulling sources from BSP and STD instead of just using the single firmware package from STM.

atsju:
I found [this](https://github.com/STMicroelectronics/stm32f429i-discovery) repository (alongside with other -discovery repositories)
It seems that repository is more up to date that the version from website you use. Note also it's quite new repo and I suppose ST is migrating all FW to github but this is not finished for every boards now.
I opened a related [issue](https://github.com/STMicroelectronics/stm32f429i-discovery/issues/1) on their repo to ask about version divergence. 

I would be better to wait for more information before merging this.

PR #133, opened 2020-06-23
Example project using uGFX and FreeRTOS for STM32F4 Discovery

alxhoff:
Requires #132


Hish15:
Hi, This seems to be too specific for this repo. What do you think @atsju?

atsju:
> Hi, This seems to be too specific for this repo. What do you think @atsju?

agreed. closing

PR #137, opened 2020-07-09
Search for startup_stm32fxxx.s locally before taking the template fil…

alxhoff:
…e from the CMSIS sources


cmake/FindCMSIS.cmake
```diff
@@ -178,10 +178,19 @@ ENDFOREACH()
...
         CMAKE_FIND_ROOT_PATH_BOTH
     )
+    IF(NOT EXISTS ${CMSIS_STARTUP_SOURCE_FILE})
+        FIND_FILE(CMSIS_STARTUP_SOURCE_FILE ${CMSIS_STARTUP_SOURCE}
+            PATH_SUFFIXES src stm32${STM32_FAMILY_LOWER} cmsis
+            HINTS ${STM32Cube_DIR}/Drivers/CMSIS/Device/ST/STM32${STM32_FAMILY}xx/Source/Templates/gcc/
+            CMAKE_FIND_ROOT_PATH_BOTH
+        )
+    ELSE()
+        MESSAGE(STATUS "CMSIS startup found locally")
```
Hish15, line 192: The message is a good idea, can't you add it in all cases and give the path. This way the user knows what's goins on

rmspacefish:
I looked into merging this for a fork gathering easy PRs, but this PR needs to be updated because the code around it changed. At some point, all capital letter CMake instructions were replaces with all lower case as well.

atsju:
@alxhoff Could you please update your branch to solve the conflict ? 
We plan to merge as much PR as possible and we would appreciate your help if you want the modification to be merged.

atsju:
after looking at  #268 it seems the modern cmake approach of v2.0 release fulfills the need. This change is no longer needed.

PR #139, opened 2020-07-25
Adding support for dual core STM32 parts on the H7 family

cwgthornton:
This is kind of gross, so I'm looking for suggestions on what to do instead. I've basically added an optional "CORE" variable to most of the find rules that will be left blank for most processors, but for the H7 it can be set to M7/M4 to target that particular device.


ObKo:
Hello there.

I was unable to find any good way to handle multi-core H7.
Your variant isn't perfect, but quite acceptable for me, I'll merge it.

Would you mind adding some information about multi-core use case to documentation or examples?

PR #141, opened 2020-08-22
feat: Created docker compile example

jorgeacortes:
Created a new example that uses a docker image containing the whole toolchain and also the cmake modules contained in this repository. Useful for CI/CD.

You can check Dockerfile and a Github actions example here: https://github.com/jorgeacortes/stm32-cmake-docker.

Blinky was taken from modern branch because Docker image has latest toolchain and seems like it's not fully compatible with current blinky at master branch.

PS: This docker image is not compatible with modern-cmake branch. Another image should be done for that instead.


atsju:
Hi @jorgeacortes 
Thank you for your contribution. I discussed with @Hish15 and we both agree this code is not relevant to this repository.

It brings nothing new to stm32-cmake implementation and is not really related to the repository.

I hope you will understand that we cannot always take everything in order to keep a maintainable quality repo.
Do not hesitate to open issue or do other PR if you see anything while using the repo.

PR #152, opened 2020-11-13
Do not compile HAL source file when only LL components are used

rimio:
Specific for users who need only LL drivers, this should prevent compile errors from missing conf header file.


ObKo:
Added better fix in aad75627d39f51753e00ffc9add0f03f95d2a211

Thanks for reporting!

Issue #71, opened 2018-02-09
Run time error on L476 due to wrong RAM size.


PhSc:
On startup, the STM32L476RG (and surely some others would) ends up in hard fault handler on first linked branch.

g-radam:
Encountered same issue for my nucleo-STM32L476RG, Rough fix - by replacing:

```
    ELSEIF(${TYPE} STREQUAL "476xx")
        SET(RAM "128K")
```
with 
```
    ELSEIF(${TYPE} STREQUAL "476xx")
        SET(RAM "96K")
```
in file gcc_stm32l4.cmake

In my search, I found that the ChibiOS's .ld for the L476 specifies 96K for general ram, and 32K for hardware parity check ram. Datasheet says "Up to 128 KB of SRAM **_including_** 32 KB _with_ hardware parity check"



ChibiOS's .ld for L476 (for reference)
```
MEMORY
{
    flash : org = 0x08000000, len = 1M
    ram0  : org = 0x20000000, len = 96k
    ram1  : org = 0x00000000, len = 0
    ram2  : org = 0x00000000, len = 0
    ram3  : org = 0x00000000, len = 0
    ram4  : org = 0x10000000, len = 32k
    ram5  : org = 0x00000000, len = 0
    ram6  : org = 0x00000000, len = 0
    ram7  : org = 0x00000000, len = 0
}

/* RAM region to be used for Main stack. This stack accommodates the processing
   of all exceptions and interrupts*/
REGION_ALIAS("MAIN_STACK_RAM", ram0);

/* RAM region to be used for the process stack. This is the stack used by
   the main() function.*/
REGION_ALIAS("PROCESS_STACK_RAM", ram0);

/* RAM region to be used for data segment.*/
REGION_ALIAS("DATA_RAM", ram0);

/* RAM region to be used for BSS segment.*/
REGION_ALIAS("BSS_RAM", ram0);

/* RAM region to be used for the default heap.*/
REGION_ALIAS("HEAP_RAM", ram0);

INCLUDE rules.ld
```

ObKo:
Closing due to https://github.com/ObKo/stm32-cmake/pull/117 being merged and made this issue outdated.

PR #162, opened 2020-12-06
Fix RAM size of the STM32L476RG

Staff-d:
Good day everyone,

this issue came up a long time ago in #71 and was never fixed.

STM advertises the STM32L476RG with a RAM size of 128K. This value, according to the [Datasheet](https://www.st.com/resource/en/datasheet/stm32l476rg.pdf), already includes the hardware parity check ram. Thus the real general purpose RAM size of the STM32L476RG is 96K.

This value is also used by the STM Cube IDE. I can confirm that anything compiled for the L476RG with the current master will throw the application right into the hard fault handler. Changing the RAM size fixes this.


Hish15:
This is a tricky one...
Are you saying that RAM size must not include the parity checked one?
Looking at the overview of the Datasheet one might think that there is 128KB of ram.
We may want to check for all families now :D

Hish15 (approved):

Staff-d:
To be honest, I am not exactly a professional when it comes to working this close to the metal, and I have no idea whether parity checked ram is usually included in the general purpose RAM. 

To shortly summarize how I got to think that this change is correct:

- I tried to get the blinky example to work on my STM32L476RG, but couldn't get any output.
- Upon searching for the error, I stumbled upon #71 and the supposed size of 96K for the RAM
- I followed a tutorial from STM for the STM32CubeIDE to get the LED to blink and compared whats different. The generated file `STM32L476RGTX_RAM.ld`, which seem to have the same purposes as the `L476RG.ld` generated by stm32-cmake, specifies a RAM size of 96K. See the excerpt below.
- Debugging the binary generated by the CubeIDE with gdb and inspecting `_estack` really gives a value of  0x20018000, which equals the the stackoffset of  0x20000000 + 96K. Inspecting the same value in the binary generated with stm32-cmake gives  0x20020000. In this case, the application seems to hang in `startup_stm32l476xx.s` on the line `ldr   sp, =_estack`, even though I am not sure if this is just a quirk of gdb.
- Finally, changing the RAM size to 96K fixes my problem and the LED blinked happily ever after

<details>
  <summary>Excerpt from STM32L476RGTX_RAM.ld</summary>

```
/* Highest address of the user mode stack */
_estack = ORIGIN(RAM) + LENGTH(RAM);	/* end of "RAM" Ram type memory */

_Min_Heap_Size = 0x200;	/* required amount of heap  */
_Min_Stack_Size = 0x400;	/* required amount of stack */

/* Memories definition */
MEMORY
{
  RAM    (xrw)    : ORIGIN = 0x20000000,   LENGTH = 96K
  RAM2    (xrw)    : ORIGIN = 0x10000000,   LENGTH = 32K
  FLASH    (rx)    : ORIGIN = 0x8000000,   LENGTH = 1024K
}
//...
```
  
</details>

atsju:
@Hish15 I had a look at datasheet. this part has 128kB of SRAM in total the problem is that they are separated in 2 banks
From [datasheet ](https://www.st.com/resource/en/datasheet/stm32l476je.pdf)section 3.15: 

- 96 Kbyte mapped at address 0x2000 0000 (SRAM1)
- 32 Kbyte located at address 0x1000 0000 with hardware parity check (SRAM2).

So 96k is correct but there should be a way to declare the other section too.

cmake/stm32/l4.cmake
```diff
@@ -16,7 +16,7 @@ set(STM32_L4_TYPE_MATCH
 set(STM32_L4_RAM_SIZES 
      40K  40K  64K  64K  64K  64K
      64K 160K 160K 160K 128K 128K
```
atsju, line 18:
```suggestion
     64K 160K 160K 160K  96K  96K

```

cmake/stm32/l4.cmake
```diff
@@ -16,7 +16,7 @@ set(STM32_L4_TYPE_MATCH
 set(STM32_L4_RAM_SIZES 
      40K  40K  64K  64K  64K  64K
      64K 160K 160K 160K 128K 128K
-    128K 128K 128K 320K 320K 320K 
+     96K 128K 128K 320K 320K 320K
```
atsju, line 19:
```suggestion
     96K  96K  96K 320K 320K 320K

```

atsju (changes requested):
I opened ALL datasheets TWICE (except 485 which seems not to exist from ST point of view, maybe a future part or one never released).

All parts have CCRAM on a second address but all except the five I corrected have also remapped this CCRAM to a second address contiguous to normal RAM. I also discovered 4A6 parts and up have a RAM3 zone wich is contiguous to the remapped CCRAM. This means all parts except the 5 corrections may work out of box with 1 big RAM zone.

Now you may make choices on how to handle this, to include or not to include in main ram, to split or not to split ...
Only 5 parts

Hish15:
Thanks @atsju
In this PR is ok to handle SRAM1 when SRAM2 is not right after it.
Another PR might handle SRAM2 in this particular case (or not 😄) 
What do you think?

Staff-d:
As @Hish15 seems to be ok with the changes from @atsju, I've included them in this PR :slightly_smiling_face:

atsju:
Hi @Staff-d 
I just wanted to inform you that I included the changes you proposed into #165. 
It goes a bit further as I checked all supported series to apply the same modifications.

Staff-d:
> Hi @Staff-d
> I just wanted to inform you that I included the changes you proposed into #165.
> It goes a bit further as I checked all supported series to apply the same modifications.

@atsju In this case, I'll close this PR and hope yours gets merged soon :slightly_smiling_face:

PR #165, opened 2020-12-14
manage CCRAM section

### LINKED ISSUE #164: CCRAM handling
This issue is linked to #162 and #71.

The original problem is the linker script absolutely not working with STM32L476RG but it goes further.

**Description of the problem for L4:**
Depending on L4 series there may be 2 or 3 RAM banks from were one is CCRAM. By default the SRAM2 (CCRAM) is mapped at `0x1000 0000` and SRAM1 at `0x2000 0000`. SRAM3 exists on some L4+ devices but forget it for now as my proposal fits to this also. 

On current implementation we have: 
- the size of SRAM1+SRAM2 is put to `RAM` section at`0x2000 0000`
- the size of SRAM2 is put to `CCRAM` section at`0x1000 0000`

This is is working on most L4 because SRAM2 is also remapped to be contiguous to SRAM1 meaning it can be accessed independently from 2 addresses. BUT if someone tries to put variables to CCRAM section then chances are he/she will overwrite his/her own variable that is in RAM.

This is not working on some devices like STM32L476 because SRAM2 is NOT remapped to be contiguous. So the stack is initialized at an address that doesn't exist on device.

Note about SRAM3: this section exists on devices where SRAM2 is remapped and is SRAM1/2/3 can all be considered contiguous.

**Proposed resolution:**
As most users want to compile with all RAM available without putting explicitly variables to section CCRAM in the code, I recommend that CCRAM section is only defined on devices where SRAM1 and SRAM2 are not contiguous. On other devices the generated linker script would put only one big RAM section and thus be easy to use 100% of the RAM for the "basic" applications.
People that explicitly need separate CCRAM section for their application would then need to write some custom linker script. This seems fair as I don't think it's most widespread case.

Note I discussed with @Hish15 about this and it seems to be best compromise. Note also this is how ST handles the thing in their examples.

**Warning this should also be checked for all other families.**
 I plan to do a PR implementing this proposal.

atsju:
Change CCRAM management to be like explained in #164 

- [x] F0 - reviewed
- [x] F1 - reviewed
- [x] F2 - reviewed
- [x] F3 - commited correction
- [x] F4 - commited correction
- [x] F7 - reviewed
- [x] G0 - reviewed and added future devices
- [x] G4 - commited correction and added new/future device support
- [x] H7 - commited correction and added new/future device support
- [x] L0 - reviewed
- [x] L1 - reviewed and minor refactor
- [x] L4 - commited correction

closes #164


cmake/stm32/g0.cmake
```diff
@@ -1,14 +1,18 @@
 set(STM32_G0_TYPES 
-    G030xx G031xx G041xx G070xx G071xx G081xx   
+    G030xx G031xx G041xx G070xx G071xx G081xx
+    G0C1xx
 )
 set(STM32_G0_TYPE_MATCH 
-    "G030.." "G031.." "G041.." "G070.." "G071.." "G081.." 
+    "G030.." "G031.." "G041.." "G070.." "G071.." "G081.."
+    "G0[BC]1.."
```
Hish15, line 7:
```suggestion
        "G030.." "G031.." "G041.." "G070.." "G071.." "G081.."
    "G0[BC]1.."
```

cmake/stm32/l4.cmake
```diff
@@ -15,17 +15,19 @@ set(STM32_L4_TYPE_MATCH
...
-      8K   8K  16K  16K  16K  16K
-     16K  32K  32K  32K  32K  32K
-     32K  32K  32K  64K  64K  64K
-     64K  64K  64K  64K  64K  64K
-     64K
+      0K   0K   0K   0K   0K   0K
+      0K   0K   0K   0K  32K  32K
+     32K  32K  32K   0K   0K   0K
+      0K   0K   0K   0K   0K   0K
+      0K
```
Hish15, line 30:
```suggestion
     0K   0K   0K   0K   0K   0K
     0K   0K   0K   0K  32K  32K
    32K  32K  32K   0K   0K   0K
     0K   0K   0K   0K   0K   0K
     0K

```

Hish15:
I tried to better align things up, check if its rendering fine

cmake/stm32/l4.cmake
atsju: this was a standard Alignement in case of 3char values. Anyway, if you consider reviewing only my spaces I consider this PR already is very high quality ;) 
What you should do is open the datasheets and check if what I do is correct :grin:

Hish15: everything counts :)

Hish15: I will not approve before checking don't worry

Hish15:
Rechecked Families By me :
- [x] L4

Hish15:
Maybe we can change the order of the device to follow ST families  e.g : 
L4 : L4x1, L4x2, L4x3, L4x5, L4x6 ?

atsju:
> Maybe we can change the order of the device to follow ST families e.g :
> L4 : L4x1, L4x2, L4x3, L4x5, L4x6 ?

Let's keep things separated. 
First fix the handling and then fix how the code is written. Maybe add this remark to #169 or create a new ticket.

ObKo:
Is this ready for merging?

atsju:
> Is this ready for merging?

@ObKo It is on my side. Maybe @Hish15 want's to finish his review ?

Hish15:
@ObKo, I'm currently having a look .
https://github.com/ObKo/stm32-cmake/pull/165#issuecomment-748461088
Thumbs up to @atsju, it is a long process to get all the data needed here.

cmake/stm32/f3.cmake
```diff
@@ -4,19 +4,19 @@ set(STM32_F3_TYPES
...
-    "303.[ED]" "318.." "328.." "334.[468]" "358.." "373.[8BC]"
+    "301.[68]" "302.[68]" "302.[BC]" "302.[DE]" "303.[68]" "303.[BC]" 
+    "303.[DE]" "318.." "328.." "334.[468]" "358.." "373.[8BC]"
     "378.." "398.."    
 )
 set(STM32_F3_RAM_SIZES 
-     16K  16K   0K  64K  16K   0K  
-     80K  16K  16K  16K  48K   0K
-     32K  80K
+     16K  16K   0K  64K  12K   0K
```
Hish15, line 12: From my understanding the third (302.[BC]) is 40Ko
image

cmake/stm32/f3.cmake
```diff
@@ -4,19 +4,19 @@ set(STM32_F3_TYPES
...
+    "301.[68]" "302.[68]" "302.[BC]" "302.[DE]" "303.[68]" "303.[BC]" 
+    "303.[DE]" "318.." "328.." "334.[468]" "358.." "373.[8BC]"
     "378.." "398.."    
 )
 set(STM32_F3_RAM_SIZES 
-     16K  16K   0K  64K  16K   0K  
-     80K  16K  16K  16K  48K   0K
-     32K  80K
+     16K  16K   0K  64K  12K   0K  
+     64K  16K  12K  12K  40K   0K
```
Hish15, line 13: The first one is 80Ko
image
4th
image
The last one :
image

cmake/stm32/f3.cmake
atsju: see #169 and look at bottom of file. I know there is zero but in fact it's implemented. You will see easier with raw files instead diff ;)

cmake/stm32/f3.cmake
```diff
@@ -4,19 +4,19 @@ set(STM32_F3_TYPES
...
+    "303.[DE]"    "318.."    "328.." "334.[468]"    "358.." "373.[8BC]"
     "378.." "398.."    
 )
 set(STM32_F3_RAM_SIZES 
-     16K  16K   0K  64K  16K   0K  
-     80K  16K  16K  16K  48K   0K
-     32K  80K
+     16K  16K   0K  64K  12K   0K  
+     64K  16K  12K  12K  40K   0K
+     32K  64K
```
Hish15, line 14: image

atsju: Yes 80 is in fact 64+16 but the 16 seems not to be aliased. So they are in CCRAM section instead. Datasheet is misleading.
I also double checked with STM32CubeF3\Drivers\CMSIS\Device\ST\STM32F3xx\Source\Templates\iar\linker\stm32f398xx_flash.icf if they did the same

Hish15:
Th F3 family is a complete mess datasheet wise. I will use `STM32CubeF3\Drivers\CMSIS\Device\ST\STM32F3xx\Source\Templates\iar\linker\st` as described in #171

atsju:
> Th F3 family is a complete mess datasheet wise. I will use `STM32CubeF3\Drivers\CMSIS\Device\ST\STM32F3xx\Source\Templates\iar\linker\st` as described in #171

Feel my pain :grin:

cmake/stm32/f3.cmake
atsju: From my understanding:
- F303.[DE] 80 = 64+16 non aliased 
- F318.. 16 this seems OK with you capture !?!?
- F318.. 16 = 12 + 4 non aliased
- F358.. 48 = 40+8 non aliased
- F373.. it's mentioned "up to" so it depends on last letter. You should look at bottom of the file for sizes

cmake/stm32/g4.cmake
```diff
@@ -1,14 +1,18 @@
...
     G431xx G441xx G471xx G473xx G483xx G474xx G484xx
+    G491xx G4A1xx
 )
 set(STM32_G4_TYPE_MATCH 
     "G431.." "G441.." "G471.." "G473.." "G483.." "G474.." "G484.."
+    "G491.." "G4A1.."
 )
 set(STM32_G4_RAM_SIZES 
      32K  32K 128K 128K 128K 128K 128K
+    112K 112K
```
Hish15, line 11: IAR template file in STM32G4_HAL gives 128K for those two which is wrong! omg

Hish15 (approved):
@ObKo, All good to me.

cmake/stm32/g4.cmake
atsju: I opened an [issue](https://github.com/STMicroelectronics/STM32CubeG4/issues/19) to ST. Good catch

Danaozhong:
Looks good!
I tested this on a STM32F303CCT6 chip. Without the change, it would immediately enter the hard fault handler, with this change it starts up normally and runs stable. At least for this chip, the RAM configuration looks good to me,

PR #174, opened 2020-12-20
.elf output file extension is managed by toolchain file

atsju:
This idea came after #163
No problem if you prefer full names. Merge only if you like it.


Hish15 (approved):
👍  I prefer this approach : a cmake target is not a elf file.
Having a .elf appended to the generated binary file is the way to go.

Hish15 (approved):

PR #182, opened 2021-01-31
print binary size

### LINKED ISSUE #166: Have binary size as output of build
In general all build IDE provide the binary size at the end.
It would be cool to add something like this code to the examples:

```
#always display the code size
ADD_CUSTOM_TARGET(do_always ALL COMMAND arm-none-eabi-size --format=berkeley "myBinary"
                                DEPENDS myBinary)
```

atsju:
closes #166 
The function is called `stm32_print_size_of_target`

I also added a small modification so that STM32_TOOLCHAIN_PATH could be set by CMAKE_C_COMPILER. As a VScode user, this is how the compiler is passed to cmake.

### DIFF
--- a/cmake/stm32/common.cmake
+++ b/cmake/stm32/common.cmake
@@ -10,11 +10,11 @@ foreach(FAMILY ${STM32_SUPPORTED_FAMILIES_LONG_NAME})
 endforeach()
 list(REMOVE_DUPLICATES STM32_SUPPORTED_FAMILIES_SHORT_NAME)
 
-if(NOT STM32_TOOLCHAIN_PATH AND NOT CMAKE_C_COMPILER)
-    set(STM32_TOOLCHAIN_PATH "/usr")
-    message(STATUS "No STM32_TOOLCHAIN_PATH specified, using default: " ${STM32_TOOLCHAIN_PATH})
-else()
-    if(NOT STM32_TOOLCHAIN_PATH)
+if(NOT STM32_TOOLCHAIN_PATH)
+    if(NOT CMAKE_C_COMPILER)
+        set(STM32_TOOLCHAIN_PATH "/usr")
+        message(STATUS "No STM32_TOOLCHAIN_PATH specified, using default: " ${STM32_TOOLCHAIN_PATH})
+    else()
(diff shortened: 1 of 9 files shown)


cmake/stm32/common.cmake
```diff
@@ -10,11 +10,17 @@ foreach(FAMILY ${STM32_SUPPORTED_FAMILIES_LONG_NAME})
...
+    message(STATUS "No STM32_TOOLCHAIN_PATH specified, using default: " ${STM32_TOOLCHAIN_PATH})
 else()
-     file(TO_CMAKE_PATH "${STM32_TOOLCHAIN_PATH}" STM32_TOOLCHAIN_PATH)
+    if(NOT STM32_TOOLCHAIN_PATH)
+        # keep only directory of compiler
+        get_filename_component(STM32_TOOLCHAIN_PATH ${CMAKE_C_COMPILER} DIRECTORY)
+        # remove the last /bin directory
+        get_filename_component(STM32_TOOLCHAIN_PATH ${STM32_TOOLCHAIN_PATH} DIRECTORY)
+    endif()
+    file(TO_CMAKE_PATH "${STM32_TOOLCHAIN_PATH}" STM32_TOOLCHAIN_PATH)
```
Hish15, line 23: The logic is confusing
```
if(not a && not b)
else() # if(a or b)
    if(not a)
    endif()
endif()
```

atsju: I did my best. What do you propose ?

Hish15: What about :
```
if(not a && not b)
elseif(not a) # the `&& b` might be a good to  
endif()
file(TO_CMAKE_PATH "${STM32_TOOLCHAIN_PATH}" STM32_TOOLCHAIN_PATH)
```

I think doing the conversion `TO_CMAKE_PATH` is good either way, so I took it out of the if blocks.
What do you think?

atsju: What's good after 3 month is that I'm stranger to my code and can judge it better. Sorry for delay :)

Hish15 (approved):

Issue #170, opened 2020-12-19
Unsupported series STM32MP1


atsju:
I think this could be integrated. At least I create the ticket for STM32 [MPU](https://www.st.com/en/microcontrollers-microprocessors/stm32-arm-cortex-mpus.html). 
Maybe someone will need it and do some pull request

ObKo:
We are talking about the built-in CM4 core, right? I don't think stm32-cmake makes sense for main A7 core.

atsju:
At first at though A7 and M4 but I had a deeper look and you are right. I don't think the A7 part make much sense in this repo.

Hish15:
Do not hesitate to close the ticket @atsju

atsju:
@Hish the CM4 core of MP1 still has (some) sense to be added into this repo. However I do not know A7 users habits and it they would be interested into this repo for CM4 co-processor

Hish15:
Ok!
For me two points :

- Things will get confusing with the way we manage multicore.
- Having just one part of a device handled by stm32-cmake seems weird

This is my 2 cents, not a definite answer. Keeping the ticket open wont cost much anyway.

sergeyblagodarnyy:
I've started this job in [my fork](https://github.com/Faramax/stm32-cmake.git) in the branch mp1.

atsju:
#283

PR #186, opened 2021-04-15
Added MP1 support

sergeyblagodarnyy:
Now it works for my project. 

For example for STM32MP157AAC chip you should do this:
```
find_package(CMSIS COMPONENTS STM32MP157AA REQUIRED)
target_link_libraries(your_executable
#   ... other libraries as is
   CMSIS::STM32::MP157AA
)

```
It configures the platform for Cortex-M4 core in the STM32MP1 family.
All chips in the MP1 family has an unique specific linker script: no flash, specific sections for IPC with Cortex-A core,... Due to this, I decided to add the original script as is instead of generated script.


sergeyblagodarnyy:
This pull request is for the #170  issue.

sergeyblagodarnyy:
@ObKo I decided to use the default linker script from ST '''stm32mp15xx_m4.ld''' It is distributed under the  BSD-3 licence, but your project uses MIT licence. I think it's not good to merge licences. Should I rewrite the script, or you know another way?

cmake/stm32/stm32mp15xx_m4.ld
```diff
@@ -0,0 +1,202 @@
+/*
```
atsju, line 1: for me no ST script should be part of this repo.
As you said there are licence problems and also version evolution will be problematic.
If the script cannot be generated by cmake then maybe we should just fall back to "user provides the script". There is an example for that I think.

sergeyblagodarnyy:
It's impossible to do cmsis fetch in the ordinary way. There is no  STMicroelectronics/cmsis_device_mp1 repository as it was for other mcus. May be STMicroelectronics/cmsis_device_f4 is appropriate for this purpose.
`
CMake Error at stm32-cmsis-mp1-subbuild/stm32-cmsis-mp1-populate-prefix/tmp/stm32-cmsis-mp1-populate-gitclone.cmake:31 (message):
  Failed to clone repository:
  'https://github.com/STMicroelectronics/cmsis_device_mp1/'
`

cmake/stm32/utilities.cmake
```diff
@@ -136,18 +136,18 @@ function(stm32_fetch_hal)
...
-        
+
         if(STM32_HAL_${FAMILY}_PATH)
             message(INFO "STM32_HAL_${FAMILY}_PATH specified, skipping fetch for ${HAL_NAME}")
             continue()
         endif()
-        
-		FetchContent_GetProperties(${HAL_NAME} POPULATED HAL_POPULATED)
+
+      FetchContent_GetProperties(${HAL_NAME} POPULATED HAL_POPULATED)
```
atsju, line 145: Please fix indentation.

cmake/stm32/utilities.cmake
```diff
@@ -42,7 +42,7 @@ endfunction()
 
 include(FetchContent)
 
-set(STM32_FETCH_FAMILIES       F0      F1      F2      F3      F4      F7      G0      G4      H7      L0      L1      L4      L5)
+set(STM32_FETCH_FAMILIES       F0      F1      F2      F3      F4      F7      G0      G4      H7      L0      L1      L4      L5 )
```
atsju, line 45: Possible to add fetch MP1 ?
+
Could you add an MP1 action for automatic build ? (We can do this in a different PR if you are not ready to do it)

cmake/stm32/stm32mp15xx_m4.ld
atsju: This file should definitely be removed and use generation done by cmake/stm32/mp1.cmake

atsju (commented):
Thank you for the update. The code seems to be in pretty good shape. 
This is not the final review but I have some preliminary remarks.

- Please update readme file to say that/how MP1 is supported (mandatory).
- Consider adding comments in the code where you think it's relevant for newcomers (recommended).

If this PR is still "work in progress" please convert the PR to draft and we will review when you are finished. If you consider it's ready for review it's OK we will review during next weeks.

sergeyblagodarnyy:
@atsju thank you for your review.
Yes, the project is not finished. Could you help me to find the button to convert PR to the draft?
I'll keep in mind your requirements.

sergeyblagodarnyy:
About  cmake/stm32/stm32mp15xx_m4.ld 
I'm agree that using it here is incorrect. But I see that it is very specific. Don't looks like others. I can put it code as is in the mp1.cmake code. Does it make sence?

atsju:
> Could you help me to find the button to convert PR to the draft?

Of course. It's [here](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request). I already converted it to draft.

> About cmake/stm32/stm32mp15xx_m4.ld
> I'm agree that using it here is incorrect. But I see that it is very specific. Don't looks like others. I can put it code as is in the mp1.cmake code. Does it make sence?

I would be better to direclty modify the linker script reference format (like for the CCRAM part that doesn't exist for all) https://github.com/ObKo/stm32-cmake/blob/master/cmake/stm32/linker_ld.cmake because MP1 is not so much different. I haven't had a deep dive into MP1 for the moment so I will let you try if you are motivated to do it. And if you need some help please reach out (I would be able to help for the embedded part, for understanding how the liker script works, what memory is for, how it is mapped, ...).

Inujel:
I am really interested in this feature. How can I contribute?

atsju:
> I am really interested in this feature. How can I contribute?

Hi @Inujel have a look at how H7 devices are added for reference.
Do the necessary changes (MP1 automated builds, remove the .ld script and modify the linker script generation) in a personnal branch and propose a pull request.
I would be great if you have your own board and run several tests to validate your code yourself.
The add of MP1 is not really an easy task, you might want to do smaller things first to understand the architecture of the repo.

sergeyblagodarnyy:
> I am really interested in this feature. How can I contribute?

You can read the discussion. I stopped on the generation the linker script. I think it's not hard to add, but I didn't have time for that.

atsju:
Done in #283

Issue #191, opened 2021-04-18
Enhance code comments


atsju:
Who has never read his own code and didn't understand it a year after ?

This repo becomes bigger, greater and more used every day and even if I'm able to write some cmake code I do not consider myself as expert. It is often quite difficult for me to understand exactly what the code does. 

I think a minimum of comment should be added and every new PR should have (in an ideal world) :

- comments for the added code
- documentation update if behavior changes

Also the original code should be commented in the futur.

Comment shall not explain what `set` does on every line but regex, general flow and "advanced" cmake features (I know this means nothing depending on people) could definitely be.

(#190 mentions this problem).

![image](https://user-images.githubusercontent.com/4628382/115136296-334c4980-a01f-11eb-9964-330c319f774d.png)

gudnimg:
I agree. Its hard to see what the code is doing sometimes :sweat_smile:

PR #192, opened 2021-04-22
Add dependency on linker script

g-arjones:


atsju (approved):
I find this really great 👍 

I would prefer more comments because I'm not expert and needed 15minutes to fully understand everything but this may be discussed in #191

cmake/stm32/common.cmake
```diff
@@ -222,6 +222,20 @@ endfunction()
...
+    if(TARGET_TYPE STREQUAL "INTERFACE_LIBRARY")
+        set(INTERFACE_PREFIX "INTERFACE_")
+    endif()
+
+    get_target_property(LINK_DEPENDS ${TARGET} ${INTERFACE_PREFIX}LINK_DEPENDS)
+    if(LINK_DEPENDS STREQUAL "LINK_DEPENDS-NOTFOUND")
+        set(LINK_DEPENDS "${SCRIPT}")
+    else()
+        list(APPEND LINK_DEPENDS "${SCRIPT}")
+    endif()
```
Hish15, line 236:
```suggestion
    if(LINK_DEPENDS)
        list(APPEND LINK_DEPENDS "${SCRIPT}")        
    else()
        set(LINK_DEPENDS "${SCRIPT}")
    endif()

```

Hish15 (changes requested):

cmake/stm32/common.cmake
g-arjones: The suggested if statement will always evaluate to `true`.

Hish15: Actually it works
https://cmake.org/cmake/help/latest/command/if.html
`False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND`

atsju: It seems that @Hish15 is right. I didn't know either.
https://cmake.org/cmake/help/latest/command/if.html

g-arjones: Interesting... And kind of awkward as well. Suggestion accepted but I feel we are wasting our time since this project is unmaintained. Especially you guys, that review all these PRs that never get merged anyway.

atsju: We are not exactly wasting our time. Obko has no time to maintain for the moment but since yesterday @Hish15 and me are officially collaborator to the project. We can and will now gradually merge all the PRs. It will take some time though.

g-arjones: Glad to hear it.

atsju (approved):

Hish15 (approved):

PR #201, opened 2021-06-06
FreeRTOS CMake update

rmspacefish:
1. Added more warnings if paths are not found
2. Added new path suffixes because of changed FreeRTOS kernel structure

New FreeRTOS kernel does not have a Source folder, see: https://github.com/FreeRTOS/FreeRTOS-Kernel


rmspacefish:
I will create a new branch for this PR because I also added more code which will be included in a separate PR

PR #203, opened 2021-06-06
Print device by family function

rmspacefish:
I added a function which simply uses the existing `stm32_get_devices_by_family` to print out
the devices without returning them. I also updated the documentation accordingly.


rmspacefish:
Replaced by #205

PR #204, opened 2021-06-06
Added example for h743zi board

rmspacefish:
I think an example for a H7 board using C++ would be useful. I struggled a bit with the M4/M7 part (well, I should have read the documentation better, I still think an example would be useful). I added that example and tested it on my H743ZI-Nucleo board.

I created two new folder `f407vg` and `h743zi`  and moved the existing example into the `f407vg` folder. It would make sense to merge this after #203 because the printout function is used (it is commented though). I added the print function because I think this is a very useful utility to include in the examples.


rmspacefish:
Replaced by #206

PR #205, opened 2021-06-08
Print device by family function

robamu:
I added a function which simply uses the existing stm32_get_devices_by_family to print out
the devices without returning them. I also updated the documentation accordingly.


atsju (commented):
I just tried the code out and
`stm32_print_devices_by_family(FAMILY STM32F4)` returns nothing. I figured out that call must be `stm32_print_devices_by_family(FAMILY F4)`. There should be a warning about unknown family in `stm32_get_devices_by_family` and maybe just some little more documentation.

Also I do not see why the function needs `FAMILY` as first keyword as it's in the name of the function. So maybe we could simplify the call.
If I omit `FAMILY` just all device get printed instead of an error that I didn't call function properly.

I think this needs a bit of rework before being merged. Maybe even `stm32_get_devices_by_family` must be reworked.

Maybe I'm a bit too extreme but we could completely remove file `devices.cmake`. This is typically something not scalable and not maintainable. And I do not see the practical use for it as we (are supposed to) support almost all STM32 devices and this list may even not fully know what we support. @Hish15 @rmspacefish what do you think of removing the file ?

rmspacefish:
I actually found this function useful to find the right target name for my specific board quickly, because the documentation does not have a similar list. I added some more documentation in the README, so I guess you mean documentation in the code? I can add that as well. I actually like the feature that it simply prints all families with non families specified.

rmspacefish:
- Warning is now emitted if the device family is invalid. Tested in example
- Added documentation in CMakeLists

README.md
```diff
@@ -184,11 +184,21 @@ CMSIS package will generate linker script for your device automatically (target
...
+To use the following functions, include the `devices.cmake` file with the following line
+
+```cmake
+include(${STM32_CMAKE_PATH}/cmake/stm32/devices.cmake)
+```
+
+where `STM32_CMAKE_PATH` is the path to the `stm32-cmake` repository.
+
+* `stm32_get_devices_by_family(DEVICES [FAMILY <family>])` - return into `DEVICES` all supported devices by family (or all devices if `<family>` is empty)
+* `stm32_print_devices_by_family([FAMILY <family>])` - Print all supported devices by family (or all devices if `<family>` is empty)
```
atsju, line 200: For me this makes more sense.
```suggestion
* `stm32_get_devices_by_family(DEVICES <family>)` - return into `DEVICES` all supported devices by family (or all devices if `<family>` is empty)
* `stm32_print_devices_by_family(<family>)` - Print all supported devices by family (or all devices if `<family>` is empty)
```
`FAMILY` is only needed when you have multiple possible keywords.

Hish15 (commented):
It does make sense to define those functions.
I find the name misleading. In fact the functions are actually `stm32_print_devices_from_family` or `stm32_print_family_devices`.
This is minor and will not block the merging of the PR, but it is my 2 cents.

README.md
rmspacefish: Hmm I thought `FAMILY` has to be specified as a keyword. Does it work without it?

atsju: No I mean you could change the implementation so that FAMILY is not used. 
It brings nothing here.

rmspacefish: I think it is okay that way. I think this was done bcause https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html is used. Using it also allows easier extensibility (e.g. you still want to parse multiple families, but you also want to pass an option or something like a filter)

atsju: > Using it also allows easier extensibility (e.g. you still want to parse multiple families, but you also want to pass an option or something like a filter)

OK I understand the point for evolution even if I don't think this function will evolve any soon.
cmake_parse_arguments is fine. I didn't know it before but it permits to do clean things.

I think if you want to have a clean function then all misuse errors should be managed
```cmake
stm32_print_devices_by_family(H7) # this shall throw error because of no FAMILY given
stm32_print_devices_by_family(FAMILY H7 L4) #this should throw error OR change FAMILY to a "multi_value_keywords"
stm32_print_devices_by_family(FAMILYH7) # this should throw error unknown keyword or something
stm32_print_devices_by_family(FAMILY) # this should throw error missing argument
```

cmake/stm32/devices.cmake
```diff
@@ -1116,6 +1116,9 @@ set(STM32_ALL_DEVICES
     WLE5JC
 )
 
+# Store a list of devices into a given DEVICES list
+# Get list of all devices for H7 family: stm32_get_devices_by_family(DEVICES FAMILY H7)
+# Print all devices: stm32_get_devices_by_family(DEVICES)
```
Hish15, line 1121:
```suggestion
# Get a list of all devices: stm32_get_devices_by_family(DEVICES)
```

cmake/stm32/devices.cmake
```diff
@@ -1125,9 +1128,15 @@ function(stm32_get_devices_by_family DEVICES)
     if(ARG_FAMILY)
         list(FILTER LIST INCLUDE REGEX "^${ARG_FAMILY}")
     endif()
+    if(NOT LIST)
```
Hish15, line 1131: This bit is weird, where does STM32_ALL_DEVICES come from? this test checks this var, not the `LIST` var.
I also strongly advice against using `LIST` as a var name. It's not self explanatory, and it is also a cmake key word!

Hish15 (changes requested):

cmake/stm32/devices.cmake
atsju: > This bit is weird, where does STM32_ALL_DEVICES come from ?

This is one of the main problems for me. It comes from here : https://github.com/ObKo/stm32-cmake/blob/master/cmake/stm32/devices.cmake
and is used only for these functions. See my very first comment in the PR where I already mentioned it and @rmspacefish seems to have an use the function.

rmspacefish:
I will introduce the safety checks mentioned by @atsju  when I find some time

rmspacefish:
I found out that the function to get the device list is used in `FindCMSIS.cmake` and in the updated `FindFreeRTOS.cmake` containing the new CMSIS support. I introduced the sanity checks and I also specified FAMILY as a multi_value_keyword. CORE is a single value keyword now because this keyword is used in `FindCMSIS.cmake` (though it appears not to be used).

atsju:
> CORE is a new single value keyword in the stm32_get_devices_by_family function now because this keyword is used in FindCMSIS.cmake (though it appears not to be used) which will lead to an empty device list because the parser then assumes CORE and ${CORE} are family keywords.

Just to mention I have seen this during #235 and I removed the CORE keyword as it is not used. You could remove it now I suppose. 

I will have a look at the rest later. Thank you for the update

rmspacefish:
Okay, `CORE` was removed

`devices.cmake` is now included like `common.cmake` as well, so both functions can be used without an extra include.

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,68 @@ set(STM32_ALL_DEVICES
...
     set(ARG_OPTIONS "")
-    set(ARG_SINGLE FAMILY)
-    set(ARG_MULTIPLE "")
+    set(ARG_SINGLE "")
+    set(ARG_MULTIPLE FAMILY)
+
+    if(ARGC GREATER_EQUAL 1)
+        if(NOT ${ARGV0} STREQUAL "STM_DEVICES")
+            message(WARNING "Passed list ${ARGV0} invalid. Still setting STM_DEVICES variable")
+        endif()
```
atsju, line 1132: this is not what the function is doing. See https://markdewing.github.io/blog/posts/notes-on-cmake/

```
stm32_get_devices_by_family(retVal FAMILY L4)
message(STATUS "retVal is ${retVal}")
message(STATUS "STM_DEVICES is ${STM_DEVICES}")
```
outputs
```
[build] CMake Warning at C:/GIT/stm32-cmake/cmake/stm32/devices.cmake:1131 (message):
[build]   Passed list retVal invalid.  Still setting STM_DEVICES variable
[build] Call Stack (most recent call first):
[build]   CMakeLists.txt:8 (stm32_get_devices_by_family)
[build] 
[build] 
[build] -- retVal is L412C8;L412CB;L412K8;L412KB;L412R8;L412RB;L412T8;L412TB;L422CB;L422KB;L422RB;L422TB;L431CB;L431CC;L431KB;L431KC;L431RB;L431RC;L431VC;L432KB;L432KC;L433CB;L433CC;L433RB;L433RC;L433VC;L442KC;L443CC;L443RC;L443VC;L451CC;L451CE;L451RC;L451RE;L451VC;L451VE;L452CC;L452CE;L452RC;L452RE;L452VC;L452VE;L462CE;L462RE;L462VE;L471QE;L471QG;L471RE;L471RG;L471VE;L471VG;L471ZE;L471ZG;L475RC;L475RE;L475RG;L475VC;L475VE;L475VG;L476JE;L476JG;L476ME;L476MG;L476QE;L476QG;L476RC;L476RE;L476RG;L476VC;L476VE;L476VG;L476ZE;L476ZG;L486JG;L486QG;L486RG;L486VG;L486ZG;L496AE;L496AG;L496QE;L496QG;L496RE;L496RG;L496VE;L496VG;L496ZE;L496ZG;L4A6AG;L4A6QG;L4A6RG;L4A6VG;L4A6ZG;L4P5AE;L4P5AG;L4P5CE;L4P5CG;L4P5QE;L4P5QG;L4P5RE;L4P5RG;L4P5VE;L4P5VG;L4P5ZE;L4P5ZG;L4Q5AG;L4Q5CG;L4Q5QG;L4Q5RG;L4Q5VG;L4Q5ZG;L4R5AG;L4R5AI;L4R5QG;L4R5QI;L4R5VG;L4R5VI;L4R5ZG;L4R5ZI;L4R7AI;L4R7VI;L4R7ZI;L4R9AG;L4R9AI;L4R9VG;L4R9VI;L4R9ZG;L4R9ZI;L4S5AI;L4S5QI;L4S5VI;L4S5ZI;L4S7AI;L4S7VI;L4S7ZI;L4S9AI;L4S9VI;L4S9ZI
[build] -- STM_DEVICES is 
```

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,68 @@ set(STM32_ALL_DEVICES
...
+    endif()
+    if(ARGC MATCHES 2)
```
atsju, line 1134: `MATCHES` is for regex. Here `EQUAL` may be more appropriate
https://cmake.org/cmake/help/latest/command/if.html#matches

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,68 @@ set(STM32_ALL_DEVICES
...
+        if(NOT ARGV1 STREQUAL "FAMILY")
+            message(WARNING "Unknown keyword: ${ARGV1}. Matching all devices..")
+        endif()
```
atsju, line 1137: I think this can be handled by `cmake_parse_arguments` see https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html
`<prefix>_KEYWORDS_MISSING_VALUES` and `<prefix>_UNPARSED_ARGUMENTS`
I think it will be safer and avoid unexpected corner cases

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,68 @@ set(STM32_ALL_DEVICES
...
-    set(LIST ${STM32_ALL_DEVICES})
+
+    # Build a list of families by filtering the whole list with the specified families
     if(ARG_FAMILY)
-        list(FILTER LIST INCLUDE REGEX "^${ARG_FAMILY}")
+        set(RESULTING_DEV_LIST "")
+        foreach(FAMILY ${ARG_FAMILY})
+            set(STM_DEVICE_LIST ${STM32_ALL_DEVICES})
+            list(FILTER STM_DEVICE_LIST INCLUDE REGEX "^${FAMILY}")
+            list(APPEND RESULTING_DEV_LIST ${STM_DEVICE_LIST})
```
atsju, line 1149: here there could be a warning when argument is not known

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,68 @@ set(STM32_ALL_DEVICES
...
+    # Some basic argument checks
+    if(ARGC MATCHES 1)
+        # FAMILY is the only keyword here
+        if(NOT ARGV STREQUAL "FAMILY")
+            message(WARNING "Unknown keyword: ${ARGV}. Matching all devices..")
+        endif()
+        if(NOT ARG_FAMILY)
+            message(STATUS "FAMILY keyword specified without any families. Matching all devices..")
+        endif()
     endif()
```
atsju, line 1179: maybe all checks are done in stm32_get_devices_by_family.

atsju (changes requested):
I need to dig further and I'm not 100% sure of how this needs to be done. If you let me some time I will try things out and propose some changes. I don't want to abuse of your time by pointing in wrong direction and needing reverse.

atsju:
I proposed changes here: https://github.com/spacefisch/stm32-cmake/pull/19
Do not hesitate to comment.

rmspacefish:
Thanks for the proposal and spotting the bugs / improvements with the CMake API. I will add some smaller tweaks/improvements as well

rmspacefish:
Some  minor improvements:

1. function name prefix not necessary for `WARNING`. CMake will already emit this information, including the line where the warning occured
2. Common parser check moved to macro, warning messages improved according to 1.
3. Bugfix in stm32_get_devices_by_family: STM_DEVICES will be set, not ${STM_DEVICES}

rmspacefish:
I added a pretty printer as well. The tests appear to be problematic now though..

rmspacefish:
I had to replace `DEVICE_FOUND` with `STM_DEVICE_FOUND`.. sometimes CMake is really confusing. Everything looks good to me now.

rmspacefish:
I did not realize the `stm32_get_devices_by_family` function is used in all the tests.. Tests are working now, I had to adapt them to the API change that the list name is `STM_DEVICES` now. I'm still considering just setting the passed list by expanding it.. what do you think?

README.md
```diff
@@ -184,11 +184,17 @@ CMSIS package will generate linker script for your device automatically (target
...
 * `stm32_print_size_of_target(<target>)` - Print the application sizes for all formats
 * `stm32_generate_binary_file(<target>)` - Generate the binary file for the given target
 * `stm32_generate_hex_file(<target>)` - Generate the hex file for the given target
 
+In the following functions, you can also specify mutiple families.
+
+* `stm32_get_devices_by_family(STM_DEVICES [FAMILY <family>])` - return into `STM_DEVICES` all
+  supported devices by family (or all devices if `<family>` is empty)
+* `stm32_print_devices_by_family([FAMILY <family>])` - Print all supported devices by family
+  (or all devices if `<family>` is empty)
```
atsju, line 196:
```suggestion
* `stm32_get_devices_by_family(STM_DEVICES [FAMILY families...])` - return into `STM_DEVICES` all
  supported devices by family (or all devices if `FAMILY` is omitted)
* `stm32_print_devices_by_family([FAMILY families...])` - Print all supported devices by family
  (or all devices if `FAMILY` is omitted)
```

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,96 @@ set(STM32_ALL_DEVICES
...
+    if(${FAMILIES} MATCHES "all")
+        message(STATUS  "Devices for all families")
+    else()
+        message(STATUS "Devices for ${FAMILIES} family")
+    endif()
+    set(TMP_LIST "")
+    foreach(STM_DEVICE ${STM_DEVICES})
+        list(APPEND TMP_LIST ${STM_DEVICE})
+        list(LENGTH TMP_LIST CURR_LEN)
+        if(CURR_LEN EQUAL 4)
```
atsju, line 1203:
```suggestion
        if(CURR_LEN EQUAL 10)
```
I don't see an usage of console less than 70chars. Avoid having too much lines ?

cmake/stm32/devices.cmake
```diff
@@ -1116,14 +1116,96 @@ set(STM32_ALL_DEVICES
...
+    endif()
+    # is populated if ARG_SINGLE or ARG_MULTIPLE is used without values
+    if(ARG_KEYWORDS_MISSING_VALUES)
+        message(FATAL_ERROR "Keyword ${ARG_KEYWORDS_MISSING_VALUES} expects values")
+    endif()
+endmacro()
+
+# Pretty printer to limit amount of list entries printed per line
+macro(stm32_pretty_print_dev_list FAMILIES STM_DEVICES)
+    if(${FAMILIES} MATCHES "all")
```
atsju, line 1194:
```suggestion
    if(${FAMILIES} STREQUAL "all")
```

Hish15 (commented):
I'm willing to do the PR, but lets stop adding commits to it first :D
Waiting the Go

atsju (approved):
@Hish15 Go
Maybe squash merge ?

Hish15 (approved):

PR #206, opened 2021-06-08
Added example for h743zi board

rmspacefish:
I think an example for a H7 board using C++ would be useful. I struggled a bit with the M4/M7 part (well, I should have read the documentation better, I still think an example would be useful). I added that example and tested it on my H743ZI-Nucleo board.

I created two new folder `f407vg` and `h743zi`  and moved the existing example into the `f407vg` folder. It would make sense to merge this after #205 (replaces #203) because the printout function is used (it is commented though). I added the print function because I think this is a very useful utility to include in the examples.


examples/freertos/f407vg/CMakeLists.txt
```diff
@@ -1,9 +1,14 @@
 cmake_minimum_required(VERSION 3.13)
-set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
+set(STM32_CMAKE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
+set(CMAKE_TOOLCHAIN_FILE ${STM32_CMAKE_PATH}/cmake/stm32_gcc.cmake)
 
 project(stm32-freertos C ASM)
 set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
 
+# Can be used to print out all devices for the F4 family
```
atsju, line 8: Note to reviewer: remember to see #205 first

examples/freertos/h743zi/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+cmake_minimum_required(VERSION 3.13)
+set(TARGET_NAME stm32-freertos)
+set(STM32_CMAKE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../..)
+
+# This must come before the project call!
+set(CMAKE_TOOLCHAIN_FILE ${STM32_CMAKE_PATH}/cmake/stm32_gcc.cmake)
+project(${TARGET_NAME} CXX C ASM)
+set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
+
+# Can be used to print out all devices for the H7 family
```
atsju, line 10: same here

examples/freertos/h743zi/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    STM32::NoSys
+)
+
+# Generate a binary file which allows flashing the STM32 board via drag-and-drop
+add_custom_command(
+	TARGET ${TARGET_NAME}
+	POST_BUILD
+	COMMAND echo Generating binary file ${CMAKE_PROJECT_NAME}.bin..
+	COMMAND ${CMAKE_OBJCOPY} -O binary ${TARGET_NAME}.elf ${TARGET_NAME}.bin
+)
```
atsju, line 47: I'm not sure this should be here. Let's keep each example minimal. But if it's useful we could either always generate elf/hex/bin OR create a function to do it depending on user needs.

atsju (commented):
Could we use same architecture as in blinky example ? I mean with same files for all boards and just 2 targets in same cmakelist.

@rmspacefish Can you please confirm: did you have a chance to try out the code on targeted board ?

examples/freertos/h743zi/CMakeLists.txt
rmspacefish: I think I added a function in another PR, but I do not use it yet. I can remove it to keep it more minimal

rmspacefish:
This code ran on my H743ZI Nucleo board successfully. I can have a look at the blinky example. However, I think it was a bit of a hassle because I had to clone two cube repositories.. I think it's more useful if the example focuses on one architecture.

atsju:
> This code ran on my H743ZI Nucleo board successfully.

Perfect, I just wanted to double check.

> I can have a look at the blinky example. However, I think it was a bit of a hassle because I had to clone two cube repositories.. I think it's more useful if the example focuses on one architecture. This example also uses C++ when compared to the other examples.

I understand the point (I just commented out the unwanted targets and I suppose I can compile just 1 target instead the all target) @Hish15 What do you think about it ?

examples/freertos/h743zi/CMakeLists.txt
rmspacefish: Will be replaced by calls to https://github.com/ObKo/stm32-cmake/pull/221 so this should probably be merged first

Hish15 (approved):
Seems good to me.

rmspacefish:
I will merge the 2 folder together again as suggested in #225

rmspacefish:
I'll replace this with an updated version using the same architecture as blinky as soon as all dependencies were merged

PR #207, opened 2021-06-08
Updated size command 

rmspacefish:
Merge after #182 , extends it with more modern CMake syntax with all lower-case letters

### DIFF
--- a/cmake/stm32/common.cmake
+++ b/cmake/stm32/common.cmake
@@ -42,11 +42,14 @@ find_program(CMAKE_SIZE NAMES ${STM32_TARGET_TRIPLET}-size PATHS ${TOOLCHAIN_BIN
 find_program(CMAKE_DEBUGGER NAMES ${STM32_TARGET_TRIPLET}-gdb PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
 find_program(CMAKE_CPPFILT NAMES ${STM32_TARGET_TRIPLET}-c++filt PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
 
-FUNCTION(stm32_print_size_of_target TARGET)
-    # this target is always considered out of date so size will always be displayed on build
-    ADD_CUSTOM_TARGET(${TARGET}_size ALL COMMAND ${CMAKE_SIZE} ${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C}
-                                                        DEPENDS ${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C})
-ENDFUNCTION()
+function(stm32_print_size_of_target TARGET)
+    add_custom_command(
+       TARGET ${TARGET}
+       POST_BUILD
+       COMMAND ${CMAKE_SIZE} ${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C}
+       COMMENT "Target Sizes: "
+    )
+endfunction()
 
 function(stm32_get_chip_type FAMILY DEVICE TYPE)
     set(INDEX 0)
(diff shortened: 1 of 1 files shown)


atsju (changes requested):
Hello @rmspacefish 
Please note #182 has been merged.
If I'm correct your commit will print size but only as post build step. If you rebuild and there is "nothing to do" then you do not print size again. The original solution will always print size even when there is nothing to do. For people developping in an IDE where the command line is erased when you hit compile it can be useful to not lose the information.

atsju (approved):

rmspacefish:
I' m getting weird errors right now, not sure if it is related. need to check this..

rmspacefish:
Hmm this is odd. I'm getting this error right now when building the `freertos` example

```
rmueller@power-pinguin:~/Software/STM32/stm32h743-cmake-minimal/stm32-cmake/examples/freertos/build$ cmake --build . -j
gmake[2]: *** No rule to make target 'stm32-freertos.elf', needed by 'CMakeFiles/stm32-freertos_size'.  Stop.
gmake[1]: *** [CMakeFiles/Makefile2:86: CMakeFiles/stm32-freertos_size.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
[  6%] Built target CMSIS_LD_F407VG
gmake: *** [Makefile:91: all] Error 2
```

atsju:
@Hish15 we need to do a CI with the actions as required with support of Obko.

rmspacefish:
Alright, I had to omit the `${CMAKE_EXECUTABLE_SUFFIX_C}` in `DEPENDS` so that the example builds for me.. The target `${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C}` does not seem to exists for CMake

cmake/stm32/common.cmake
```diff
@@ -42,11 +42,13 @@ find_program(CMAKE_SIZE NAMES ${STM32_TARGET_TRIPLET}-size PATHS ${TOOLCHAIN_BIN
 find_program(CMAKE_DEBUGGER NAMES ${STM32_TARGET_TRIPLET}-gdb PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
 find_program(CMAKE_CPPFILT NAMES ${STM32_TARGET_TRIPLET}-c++filt PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
 
-FUNCTION(stm32_print_size_of_target TARGET)
-    # this target is always considered out of date so size will always be displayed on build
-    ADD_CUSTOM_TARGET(${TARGET}_size ALL COMMAND ${CMAKE_SIZE} ${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C}
-                                                        DEPENDS ${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C})
-ENDFUNCTION()
+function(stm32_print_size_of_target TARGET)
+    add_custom_target(always_display_size
```
atsju, line 46: I just wanted to try on `blinky `example and now remember that I used `${TARGET}_size` because in case of multiple targets you are redefining it. You need to create an unique _size target per target.
```suggestion
    add_custom_target(${TARGET}_always_display_size
```

atsju (changes requested):
I tested with `DEPENDS ${TARGET}` and with `DEPENDS ${TARGET}${CMAKE_EXECUTABLE_SUFFIX_C}`

Both are working on blinky example. I didn't push investigation further

cmake/stm32/common.cmake
rmspacefish: That makes sense. I added the change

atsju (approved):

Hish15 (approved):

PR #221, opened 2021-06-27
added binary and hex generation functions

rmspacefish:
Functions to generate binary file and hex file


gudnimg:
I think it would be beneficial to add `BYPRODUCTS ${TARGET}.hex` after `COMMAND`.

They mention this in the  docs:
> The Makefile Generators will remove BYPRODUCTS and other GENERATED files during make clean.

I believe the same applies to `ninja clean`.

atsju:
Please update the readme file also.

atsju (approved):

rmspacefish:
The `${CMAKE_EXECUTABLE_SUFFIX_C}` is needed because the file might or might not have a suffix and the `arm-none-eabi-size` command needs the exact binary name. If the suffix is empty, it's not an issue because the resulting expansion will be empty too

gudnimg:
> The `${CMAKE_EXECUTABLE_SUFFIX_C}` is needed because the file might or might not have a suffix and the `arm-none-eabi-size` command needs the exact executable name. If the suffix is empty, it's not an issue because the resulting expansion will be empty too.

I was not aware of such a limitation. This has worked for me before:

```cmake
# Print size information after build
add_custom_command(TARGET ${target} POST_BUILD
    COMMAND ${CMAKE_SIZE} "${target}"
)
```

`CMAKE_OBJCOPY` is `arm-none-eabi-objcopy`. I know it doesn't need an extension because it worked for me in another project. My executable had no extension such as `.exe`, `.elf`.

I don't quite see yet why `${CMAKE_EXECUTABLE_SUFFIX_C}` is needed with `arm-none-eabi-objcopy` or how that relates to `arm-none-eabi-size`

atsju:
@gudnimg You are right, it's related to #174.
Because every file is formated in some way and I find it good to have the corresponding extension

Hish15 (approved):

PR #223, opened 2021-07-02
Added Segger RTT example

stawiski:
Hi,

I added an example of SEGGER RTT through JLink debug probe.
This example should seamlessly run on NUCLEO-G071RB Nucleo Evaluation board.

Requires:
- JLink debug probe or converting the Nucleo's built-in STLink to JLink
- SEGGER RTT Viewer or Ozone debugger to show the `printf` output

Cheers


atsju:
Hi @stawiski 
Thank you very much for your contribution. I am a big fan of Segger RTT myslef but I discussed with @Hish15 and we both agree this code is not relevant to this repository.

It brings nothing new to stm32-cmake implementation and would be quite easy to configure so that it compiles for any user needing it. Also the RTT sources would not be kept up to date with latest sources from Segger.

I hope you will understand that we cannot always take everything in order to keep a maintainable quality repo.
Do not hesitate to open issue or do other PR if you see anything while using the repo.

PR #225, opened 2021-07-02
Blinky example minor improvement

rmspacefish:
I updated `blinky` example to allow choosing the architecture via CMake options, with F4 being the default. I think this is a cleaner solution than commenting out the unneeded code or cloning 3 Cube repositories. What do you think?


examples/blinky/CMakeLists.txt
```diff
@@ -4,9 +4,25 @@ set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake
 project(stm32-blinky C ASM)
 set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
 
-find_package(CMSIS COMPONENTS STM32L0 STM32F1 STM32F4 REQUIRED)
-find_package(HAL COMPONENTS STM32L0 STM32F1 STM32F4 RCC GPIO CORTEX REQUIRED)
-# Find all drivers:
+# F4 example enabled by default
```
atsju, line 7:
```suggestion
# Configure here which STM32 target(s) to build
# F4 example enabled by default
```

examples/blinky/CMakeLists.txt
```diff
@@ -4,9 +4,25 @@ set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake
...
+option(F4_EXAMPLE "Compile F4 example" ON)
+option(F1_EXAMPLE "Compile F1 example" OFF)
+option(L0_EXAMPLE "Compile L0 example" OFF)
+
+find_package(CMSIS COMPONENTS STM32F4 REQUIRED)
+find_package(HAL COMPONENTS STM32F4 RCC GPIO CORTEX REQUIRED)
+
+if(L0_EXAMPLE)
+    find_package(CMSIS COMPONENTS STM32L0 REQUIRED)
+    find_package(HAL COMPONENTS STM32L0 REQUIRED)
```
atsju, line 17: I  didn't test but shouldn't it be this ?
```suggestion
    find_package(HAL COMPONENTS STM32L0 RCC GPIO CORTEX REQUIRED)
```

examples/blinky/CMakeLists.txt
```diff
@@ -4,9 +4,25 @@ set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake
 project(stm32-blinky C ASM)
 set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
 
-find_package(CMSIS COMPONENTS STM32L0 STM32F1 STM32F4 REQUIRED)
-find_package(HAL COMPONENTS STM32L0 STM32F1 STM32F4 RCC GPIO CORTEX REQUIRED)
-# Find all drivers:
+# F4 example enabled by default
+option(F4_EXAMPLE "Compile F4 example" ON)
```
atsju, line 8: It's strange to  have an option with no effect. Add the required `if(F4_EXAMPLE)`

atsju (changes requested):
The idea seams good. This method could be used in #206 ? @Hish15 what do you think ?

examples/blinky/CMakeLists.txt
Hish15: selecting no component will look up all components, so this should work.

atsju: But is it what we want ? and it's not same as L4 lines.
If we add a comment it's OK for me

examples/blinky/CMakeLists.txt
Hish15: I agree, make it ON by default, but allow to turn it OFF using the adequate if(F4_EXAMPLE)

Hish15 (changes requested):
Hi!
I love what you have done here 👍
some minor changes : 
The scopes are quite large with cmake, so I would use `BLINKY_EXAMPLE_F4` kind of name instead of just `F4_EXAMPLE`. This will limit possible collisions.

examples/blinky/CMakeLists.txt
rmspacefish: Updated

examples/blinky/CMakeLists.txt
rmspacefish: Updated to use `find_package` with one call by filling component lists

examples/blinky/CMakeLists.txt
rmspacefish: Implemented

rmspacefish: Using it in #206 seems a good idea and I actually have a F4 board to test this as well. I might supply it in a separate PR

Hish15 (approved):

atsju (approved):

examples/blinky/CMakeLists.txt
atsju: @rmspacefish yes please go ahead. This is a clean way to do it.

PR #229, opened 2021-07-13
utilities.cmake update default build flags

### LINKED ISSUE #228: utilities.cmake enforces compiler flags and optimization levels
Hello,

I recently integrated `stm32-cmake` into our project and stumbled on some issue related to compiler flags and options.

The `utilities.cmake` enforces three compiler flags/optimization, which are not ABI flags (e.g. `-mcpu=cortex-m7` or `-mthumb`) and not strictly necessary. It probably would be better to let the user device which optimization flags to take and where to opt out.

The three optimizations options in questions are:

- `-ffast-math`
- `-fno-builtin` . This flag has actually caused compilation error in our project because some math funtions could not be used in a `constexpr` context for `static_assert` anymore
- `fno-strict-aliasing` . Is usually enabled on `-Os`, `-O2` and `-O3`. I have compiled without this without issues

Everything can be looked up here: https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html
If some of those options are strongly recommended in an embedded context, I' d still prefer to have an `option` for them with the default value `ON` so it is possible to disable them from an upper CMakeLists.txt file. Another option would be to remove them completely, leaving compiler flag selection to the user.

The same cmake file also enforces optimization levels: It will cause CMake to use `-Os` if the `Release` configuration is used.
I think this is also something the user should decide. I compiled our application with the default setting for `Release` (`-O3`) and it worked without issues.

rmspacefish:
Closes #228 . This makes some optimization flags optional. There is also the option to remove them altogether. If they are not removed, I might add some documentation.

I also removed the enforcement of optimization flags for build configurations

### DIFF
--- a/cmake/stm32/utilities.cmake
+++ b/cmake/stm32/utilities.cmake
@@ -13,7 +13,7 @@ function(stm32_util_create_family_targets FAMILY)
 
     if(NOT (TARGET STM32::${FAMILY}${CORE_C}))
         set(STM32_COMPILE_OPTIONS
-            -mthumb -mabi=aapcs -Wall -ffunction-sections -fdata-sections
+            -mthumb -Wall -ffunction-sections -fdata-sections
         )
         if(STM32_ENABLE_FAST_MATH)
             list(APPEND STM32_COMPILE_OPTIONS -ffast-math)
--- a/cmake/stm32/utilities.cmake
+++ b/cmake/stm32/utilities.cmake
@@ -14,7 +14,7 @@ function(stm32_util_create_family_targets FAMILY)
     if(NOT (TARGET STM32::${FAMILY}${CORE_C}))
         add_library(STM32::${FAMILY}${CORE_C} INTERFACE IMPORTED)
         # Set compiler flags for target
-        # -Wall: all warning activated
+        # -Wall: all warnings activated
         # -ffunction-sections -fdata-sections: remove unused code
         target_compile_options(STM32::${FAMILY}${CORE_C} INTERFACE 
             --sysroot="${TOOLCHAIN_SYSROOT}"
--- a/cmake/stm32/utilities.cmake
+++ b/cmake/stm32/utilities.cmake
@@ -11,19 +11,32 @@ function(stm32_util_create_family_targets FAMILY)
         message(FATAL_ERROR "Expected at most one core for family ${FAMILY}: ${CORES}")
     endif()
 
+    option(STM32_ENABLE_FAST_MATH "Add -ffast-math compiler flag" OFF)
+    option(STM32_NO_BUILTIN "Add -fno-builtin compiler flag" OFF)
+    option(STM32_NO_STRICT_ALIASING "Add -fno-strict-aliasing compiler flag" OFF)
+
     if(NOT (TARGET STM32::${FAMILY}${CORE_C}))
+        set(STM32_COMPILE_OPTIONS
+            -mthumb -mabi=aapcs -Wall -ffunction-sections -fdata-sections
+        )
+        if(STM32_ENABLE_FAST_MATH)
+            list(APPEND STM32_COMPILE_OPTIONS -ffast-math)
+        endif()
+        if(STM32_NO_BUILTIN)
+            list(APPEND STM32_COMPILE_OPTIONS -fno-builtin)
+        endif()
+        if(STM32_NO_STRICT_ALIASING)
+            list(APPEND STM32_COMPILE_OPTIONS -fno-strict-aliasing)
+        endif()
+
         add_library(STM32::${FAMILY}${CORE_C} INTERFACE IMPORTED)
         target_compile_options(STM32::${FAMILY}${CORE_C} INTERFACE 
             --sysroot="${TOOLCHAIN_SYSROOT}"
-            -mthumb -mabi=aapcs -Wall -ffunction-sections -fdata-sections -fno-strict-aliasing -fno-builtin -ffast-math
-            $<$<CONFIG:Debug>:-Og>
-            $<$<CONFIG:Release>:-Os>
+            ${STM32_COMPILE_OPTIONS}
         )
         target_link_options(STM32::${FAMILY}${CORE_C} INTERFACE 
             --sysroot="${TOOLCHAIN_SYSROOT}"
             -mthumb -mabi=aapcs -Wl,--gc-sections
-            $<$<CONFIG:Debug>:-Og>
-            $<$<CONFIG:Release>:-Os -s>
         )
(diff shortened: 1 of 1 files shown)


cmake/stm32/utilities.cmake
```diff
@@ -11,19 +11,32 @@ function(stm32_util_create_family_targets FAMILY)
         message(FATAL_ERROR "Expected at most one core for family ${FAMILY}: ${CORES}")
     endif()
 
+    option(STM32_ENABLE_FAST_MATH "Add -ffast-math compiler flag" OFF)
+    option(STM32_NO_BUILTIN "Add -fno-builtin compiler flag" OFF)
+    option(STM32_NO_STRICT_ALIASING "Add -fno-strict-aliasing compiler flag" OFF)
+
```
atsju, line 17: Is there a way for user to add his own (any) options ? 
If yes I would consider removing the `option` and just let him manage in his own way. I mean these options are not more important than other that the user might need.
```suggestion

```

cmake/stm32/utilities.cmake
```diff
@@ -11,19 +11,32 @@ function(stm32_util_create_family_targets FAMILY)
         message(FATAL_ERROR "Expected at most one core for family ${FAMILY}: ${CORES}")
     endif()
 
+    option(STM32_ENABLE_FAST_MATH "Add -ffast-math compiler flag" OFF)
+    option(STM32_NO_BUILTIN "Add -fno-builtin compiler flag" OFF)
+    option(STM32_NO_STRICT_ALIASING "Add -fno-strict-aliasing compiler flag" OFF)
+
     if(NOT (TARGET STM32::${FAMILY}${CORE_C}))
+        set(STM32_COMPILE_OPTIONS
+            -mthumb -mabi=aapcs -Wall -ffunction-sections -fdata-sections
```
atsju, line 20: I'm usure about what does `-mabi=aapcs` do ?
I never used it before maybe it could be removed also
```suggestion
            -mthumb -Wall -ffunction-sections -fdata-sections
```

atsju (commented):
You spotted good points in  #228 👍 
I didn't knew `-ffast-math` but this seems really dangerous to enable by default

rmspacefish (commented):
- Removed `-mabi=aapcs`, user can specify ABI if a special one is required
- Removed options, user can supply optimization flags when required

cmake/stm32/utilities.cmake
```diff
@@ -12,18 +12,20 @@ function(stm32_util_create_family_targets FAMILY)
...
+            -mthumb -Wall -ffunction-sections -fdata-sections
+        )
+
         add_library(STM32::${FAMILY}${CORE_C} INTERFACE IMPORTED)
         target_compile_options(STM32::${FAMILY}${CORE_C} INTERFACE 
             --sysroot="${TOOLCHAIN_SYSROOT}"
-            -mthumb -mabi=aapcs -Wall -ffunction-sections -fdata-sections -fno-strict-aliasing -fno-builtin -ffast-math
-            $<$<CONFIG:Debug>:-Og>
-            $<$<CONFIG:Release>:-Os>
+            ${STM32_COMPILE_OPTIONS}
```
atsju, line 23:
```suggestion
            -mthumb -Wall -ffunction-sections -fdata-sections

```

cmake/stm32/utilities.cmake
```diff
@@ -12,18 +12,20 @@ function(stm32_util_create_family_targets FAMILY)
     endif()
 
     if(NOT (TARGET STM32::${FAMILY}${CORE_C}))
+        # Remove unused code, generate thumb code and enable all warnings
+        set(STM32_COMPILE_OPTIONS
+            -mthumb -Wall -ffunction-sections -fdata-sections
+        )
+
```
atsju, line 19: The intermediate variable has no use now that `OPTION` are gone
```suggestion

```

cmake/stm32/utilities.cmake
```diff
@@ -12,18 +12,20 @@ function(stm32_util_create_family_targets FAMILY)
     endif()
 
     if(NOT (TARGET STM32::${FAMILY}${CORE_C}))
+        # Remove unused code, generate thumb code and enable all warnings
+        set(STM32_COMPILE_OPTIONS
+            -mthumb -Wall -ffunction-sections -fdata-sections
+        )
+
         add_library(STM32::${FAMILY}${CORE_C} INTERFACE IMPORTED)
         target_compile_options(STM32::${FAMILY}${CORE_C} INTERFACE
```
atsju, line 21:
```suggestion
        # Remove unused code, generate thumb code and enable all warnings
        target_compile_options(STM32::${FAMILY}${CORE_C} INTERFACE 

```

Hish15: @atsju This comment will be wrong when we will update any of the flags. I see two fixes:
 1. Make the comment more basic "sets the compiler flags for the target blablabla"
 2.  List the current flags alongside a brief description of what each does.

I'm more for the solution 1 here.

cmake/stm32/utilities.cmake
```diff
@@ -12,18 +12,20 @@ function(stm32_util_create_family_targets FAMILY)
...
-            -mthumb -mabi=aapcs -Wall -ffunction-sections -fdata-sections -fno-strict-aliasing -fno-builtin -ffast-math
-            $<$<CONFIG:Debug>:-Og>
-            $<$<CONFIG:Release>:-Os>
+            ${STM32_COMPILE_OPTIONS}
         )
+        # Remove unused code, generate thumb code
         target_link_options(STM32::${FAMILY}${CORE_C} INTERFACE 
             --sysroot="${TOOLCHAIN_SYSROOT}"
-            -mthumb -mabi=aapcs -Wl,--gc-sections
-            $<$<CONFIG:Debug>:-Og>
```
Hish15, line 25: I like the -Og for debug builds, why is that bad?

atsju: It's bad because it overrides Cmake default behavior and default behavior is fine. 
I tested Debug(-g)/Release(-O3)/MinSizeRel(-Os)/ReleaseWithDebInfo(-O2 -g)

cmake/stm32/utilities.cmake
atsju: Updated

rmspacefish: Done

cmake/stm32/utilities.cmake
rmspacefish: Done

cmake/stm32/utilities.cmake
rmspacefish: Removed

atsju (approved):
@Hish15 please review. Squash and merge is appropriate here.

PR #218, opened 2021-06-22
Add warning message if gdb is not installed/found.

gudnimg:
Resolves #211

Issue #211, opened 2021-06-13
Bump minimum CMake version to 3.18 to improve find_program()?


gudnimg:
With CMake 3.18 there is a `REQUIRED` option for `find_program()`

Example: If the `arm-none-eabi-gcc` is not found, the CMake will stop processing and create an error message.

**CMake >= 3.18**:
```cmake
find_program(CMAKE_C_COMPILER NAMES ${STM32_TARGET_TRIPLET}-gcc PATHS ${TOOLCHAIN_BIN_PATH} REQUIRED)
```
**CMake < 3.18**:
```cmake
find_program(CMAKE_C_COMPILER NAMES ${STM32_TARGET_TRIPLET}-gcc PATHS ${TOOLCHAIN_BIN_PATH})
```

rmspacefish:
This might also be worth an own issue.. There are also some calls to find utilities in the `common.cmake` file:

```cmake
find_program(CMAKE_OBJCOPY NAMES ${STM32_TARGET_TRIPLET}-objcopy PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_OBJDUMP NAMES ${STM32_TARGET_TRIPLET}-objdump PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_SIZE NAMES ${STM32_TARGET_TRIPLET}-size PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_DEBUGGER NAMES ${STM32_TARGET_TRIPLET}-gdb PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_CPPFILT NAMES ${STM32_TARGET_TRIPLET}-c++filt PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
```

I am used to add the toolchain binary path to the system path and then simply look for `arm-none-eabi-XYZ` instead of setting a specific variable like `STM32_TOOLCHAIN_PATH` . Then again, the later one allows multiple toolchain (although I guess this is not really relevant for most users)..  I don't know the exact reasoning behind the `NO_DEFAULT_PATH` option, but maybe it might be a good idea to omit that option. Those utilities are not really required, but every toolchain should have them, so maybe `REQUIRED` can be added there as well.

gudnimg:
[...]

PR #233, opened 2021-07-15
Toolchain file update

### LINKED ISSUE #216: Seach for optional modules in environment as well
There are also some calls to find utilities in the `common.cmake` file to find utilities like `arm-none-eabi-objcopy` or `arm-none-eabi-size`:

```cmake
find_program(CMAKE_OBJCOPY NAMES ${STM32_TARGET_TRIPLET}-objcopy PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_OBJDUMP NAMES ${STM32_TARGET_TRIPLET}-objdump PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_SIZE NAMES ${STM32_TARGET_TRIPLET}-size PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_DEBUGGER NAMES ${STM32_TARGET_TRIPLET}-gdb PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_CPPFILT NAMES ${STM32_TARGET_TRIPLET}-c++filt PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
```

For the necessary toolchain components, adding the binary folder to the environment is already sufficient.
I think this would be good here as well so it behaves consistently. I think removing `NO_DEFAULT_PATH` is sufficient to get this
behaviour.

rmspacefish:
1. Use `STM32_TOOLCHAIN_PATH` from environment if specified as environmental variable
2. Find tools like objcopy and size in default paths as well, which allows
   simply adding the binary folder of the toolchain to the path to find them.
   This also allows more consistent behaviour, because this already works
   for <cross-compile>-gcc/g++

closes #216


atsju:
For information you might have a look at #218 + #211

rmspacefish:
Ok. Maybe we can merge it first? It looks finished, but I have not read everything yet

atsju:
I'm at work now. But looks fine, I will test ASAP and it can be merged regardless of other PR. As I said it was only for information.

Hish15 (changes requested):
Here setting `STM32_TOOLCHAIN_PATH` in a env var, will make setting the value directly impossible. I think we want to check if the var is not set (using CLI or in the cmakelist) before using ENV var.

cmake/stm32/common.cmake
```diff
@@ -10,6 +10,12 @@ foreach(FAMILY ${STM32_SUPPORTED_FAMILIES_LONG_NAME})
 endforeach()
 list(REMOVE_DUPLICATES STM32_SUPPORTED_FAMILIES_SHORT_NAME)
 
+if(DEFINED ENV{STM32_TOOLCHAIN_PATH})
+    message(STATUS "Detected toolchain path STM32_TOOLCHAIN_PATH in environmental variables: ")
+    message(STATUS "$ENV{STM32_TOOLCHAIN_PATH}")
+    set(STM32_TOOLCHAIN_PATH $ENV{STM32_TOOLCHAIN_PATH})
+endif()
+
 if(NOT STM32_TOOLCHAIN_PATH)
```
atsju, line 19:
```suggestion
if(NOT STM32_TOOLCHAIN_PATH)
    if(DEFINED ENV{STM32_TOOLCHAIN_PATH})
        message(STATUS "Detected toolchain path STM32_TOOLCHAIN_PATH in environmental variables: ")
        message(STATUS "$ENV{STM32_TOOLCHAIN_PATH}")
        set(STM32_TOOLCHAIN_PATH $ENV{STM32_TOOLCHAIN_PATH})
    endif()
```
@Hish15  is this what you recommend ? Would make sense.

Hish15: looks like it yeah 😄

atsju (approved):

Hish15 (approved):

Issue #241, opened 2021-07-22
Blinky Example doesn't compile due to lack of LL Components


BenArtes:
**Describe the bug**
When attempting to build the `blinky` example I receive the following error.

```bash
-- No STM32_TOOLCHAIN_PATH specified, using default: /usr
-- No STM32_TARGET_TRIPLET specified, using default: arm-none-eabi
-- The C compiler identification is GNU 6.3.1
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/arm-none-eabi-gcc
-- Check for working C compiler: /usr/bin/arm-none-eabi-gcc
-- Check for working C compiler: /usr/bin/arm-none-eabi-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Neither STM32_CUBE_F4_PATH nor STM32_CMSIS_F4_PATH specified using default  STM32_CUBE_F4_PATH: /opt/STM32CubeF4
-- Found CMSIS: /opt/STM32CubeF4/Drivers/CMSIS/Include;/opt/STM32CubeF4/Drivers/CMSIS/Device/ST/STM32F4xx/Include (found version "5.4.0") found components:  STM32F4 
CMake Error at /home/terminus/stm32-cmake/cmake/FindHAL.cmake:220 (list):
  list sub-command REMOVE_DUPLICATES requires list to be present.
Call Stack (most recent call first):
  CMakeLists.txt:31 (find_package)

```

I presume that the lack of LL drivers in the blinky example causes `  list(REMOVE_DUPLICATES HAL_FIND_COMPONENTS_DRIVERS_LL)` to fail.

I commented out the code as a test and then attempted to build blinky-ll and ran into similar errors related to `HAL_FIND_COMPONENTS_DRIVERS`.

I modified FindHAL.cmake to have checks surrounding the relevant blocks as below and all examples built correctly.

If this is a correct / desirable change I'm happy to create a pull request.

see #242

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Compiler: arm-none-eabi-gcc (15:6.3.1+svn253039-1build1) 6.3.1 20170620
 - stm32-cmake: https://github.com/ObKo/stm32-cmake/commit/cebbda690db674885cf0dca2a90dd8f4426d40d0

atsju:
Thank you for reporting this issue. 
[...]

PR #242, opened 2021-07-22
Checks if HAL / LL components are defined in FindHAL.cmake before using.

BenArtes:
Potential fix for missing HAL / LL components as described in https://github.com/ObKo/stm32-cmake/issues/241.

Issue is currently for easy diff comparison.


BenArtes:
As per Issue #241, minimum CMake version will be bumped to 3.16 instead.

New PR: https://github.com/ObKo/stm32-cmake/pull/245

PR #261, opened 2021-08-08
added gitignore for examples folder

robamu:
This first version ignores folders starting with build.
Most CMake projects create the build files inside a build folder
out of the source tree, so this makes working with git in the examples
more convenient


Hish15:
Why limit this to the example folder? we do a have a `test/` folder too.
I think it's better to put this all the way to the top.
What do you think @robamu and @atsju ?

robamu:
I think it is good. As long as their are no folders prefixed with build, this should be fine.

atsju:
`.gitignore` would only be a convenience file for us developpers. Note that `build` folder is a default name and any developper could build to a different place or rename this folder.

I have no strong opinion about adding it or not. 

If we add it please put it at repo root. And use `build/` instead `build` as it will target folders only. You can force push as this is a one liner.

robamu:
Done

Hish15 (approved):

PR #262, opened 2021-08-08
README tweaks

robamu:
1. Additional example for HAL usage with the H7 family using two cores
2. HTML ID achors for headings added
3. Some minor formatting improvements and line breaks
4. Additional usage example for setting the toolchain file

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,17 +88,22 @@ $env:TARGET_TRIPLET = arm-none-eabi
 $env:STM32_CUBE_<FAMILY>_PATH="<PathToCubeRoot>"
 ```
 
-## Common usage
+## <a id="cmsis"></a> Common usage
 
 First thing that you need to do after toolchain configuration in your `CMakeLists.txt` script is to find CMSIS package:
 ```cmake
 find_package(CMSIS [CMSIS_version] COMPONENTS STM32F4 REQUIRED)
 ```
-You can specify STM32 family or even specific device (`STM32F407VG`) in `COMPONENTS` or omit `COMPONENTS` totally - in that case stm32-cmake will find ALL sources for ALL families and ALL chips (you'll need ALL STM32Cube packages somewhere).
+You can specify STM32 family or even specific device (`STM32F407VG`) in `COMPONENTS` or omit
+`COMPONENTS` totally - in that case stm32-cmake will find ALL sources for ALL families and
+ALL chips (you'll need ALL STM32Cube packages somewhere).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,16 +187,33 @@ target_link_libraries(stm32-blinky-f4
...
+Here is another usage for a H7 device with the M7 core:
 
+```cmake
+target_link_libraries(${TARGET_NAME} PRIVATE
+    HAL::STM32::H7::M7::RCC
+    HAL::STM32::H7::M7::GPIO
+    HAL::STM32::H7::M7::CORTEX
+    CMSIS::STM32::H743ZI::M7
+    STM32::NoSys
+)
```
atsju, line 199: for all "examples", shouldn't we just link to the example folder ? I mean the readme is here to explain the targets but we have dedicated examples so we should not duplicate the information here. 
@Hish15  @robamu what do you think ?

robamu: I think it's okay if the README has a few examples as well to have a general idea how the targets are used. I added that example because it is one that uses two cores.

atsju: Sorry for delay, we have had less time to dedicate to this repo.
In my experience duplicating too much information in the documentation often ends with documentation not up to date when the code evolves. This is why I think a link to example is better than a copy.

@Hish15  any thoughts about this ?

robamu: Sounds good. I'll look into it

robamu: I specifically added the dual core example to have at least one simple example for a dual core.
Maybe it would be a good idea to add the H7M7 example to the blinky application?

Hish15:
Hi,

I personally think that splitting the line like this does not bring any value for a maintainer (The line break changes does not github rendering). Maybe It does, but I cannot see it.
Some cleaning is needed see #270, but none it's actually done here. I think it's better to close this PR without merging it.

robamu:
Do you want to discard all the suggested changes? The anchors are useful to add cross-references and I think some of the additions increase the usability of the project (e.g. example of hardcoding the toolchain file to prevent setting the same one all the time).

About the line breaks: You don't see them in github, but I think it's nicer for someone who reads / edits them in a text editor

Hish15:
You are right on **the lines breaks**, as I said myself it  does not change the generated html file. We can keep them.

The Usage section at the beginning of the file state that you can add multiple vars in different ways, and links to the exemples are provided. CMAKE_TOOLCHAIN_FILE is not in the list of those vars! It's assumed that people will know.
**Lets add there instead of at the end.**

robamu:
I moved the explanation to the top

Issue #251, opened 2021-07-28
Could no set path of compiler


zisi:
**Describe the bug**
I set the compiler by using:

```
set(STM32_TARGET_TRIPLET "arm-none-eabi")
set(STM32_TOOLCHAIN_PATH "${CMAKE_CURRENT_SOURCE_DIR}/tools/gcc-arm-none-eabi-latest")
``` 
but the stm32-cmake uses the usr path to find compiler. Finally i change the stm32_gcc.cmake

```
find_program(CMAKE_C_COMPILER NAMES ${STM32_TARGET_TRIPLET}-gcc PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_CXX_COMPILER NAMES ${STM32_TARGET_TRIPLET}-g++ PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
find_program(CMAKE_ASM_COMPILER NAMES ${STM32_TARGET_TRIPLET}-gcc PATHS ${TOOLCHAIN_BIN_PATH} NO_DEFAULT_PATH)
``` 
in order do not search in default path (usr).

**Sources to reproduce**
-

**Steps to reproduce**
-

**Expected behavior**
-

**Screenshots**
-

**Environment (please complete the following information):**
 - OS: kubuntu 20.004
 - Compiler: -
 - stm32-cmake: e9c22b93fed1f102fbe02a3ccd73737846fc9c99
 - cmake: 3.21
 - HAL/cube/CMSIS: -
 - 
**Additional context**
-

atsju:
Thank you for reporting the issue. 
This might be a side effect of #233 
Unfortunately I'm not 100% understanding the effect of `NO_DEFAULT_PATH`.
If you understood the root cause, could you please elaborate a bit more ?

zisi:
As i understand by using `NO_DEFAULT_PATH` the searching path is restricted to `${TOOLCHAIN_BIN_PATH}`.

Reference: https://cmake.org/cmake/help/latest/command/find_program.html

atsju:
> As i understand by using `NO_DEFAULT_PATH` the searching path is restricted to `${TOOLCHAIN_BIN_PATH}`.

What exactly is the problem with this for you ? In the order of evaluation, `HINTS` is in position 4. What makes you find a wrong gcc (if this is your issue) ?
[...]

PR #265, opened 2021-08-10
No default path option for finding toolchain

robamu:
This commits adds the support to skip the default CMake paths
when finding the cross-compile toolchain binaries.

This was suggested in #251 .


atsju:
@robamu thank you for this PR. Are you able to reproduce the original issue ? 
if so, please just try `set(STM32_TOOLCHAIN_PATH "${CMAKE_CURRENT_SOURCE_DIR}/tools/gcc-arm-none-eabi-latest" CACHE)`

Notice the `CACHE`. Idea comes from [documentation](https://cmake.org/cmake/help/latest/command/find_program.html)

We don't want to overcomplicate things in the code and `NO_DEFAULT_PATH` doesn't seem to be the correct practice here. This use case is not rocket science so there must be an easy way to solve.

robamu:
I do not have the toolchain installed in one of the default locations specified by CMake. What I can do is to try to only set `STM32_TOOLCHAIN_PATH` and add a different path containing the same toolchain binaries to my system path and then check whether the toolchain at `STM32_TOOLCHAIN_PATH` is picked. I was also hoping that @zisi can test whether this solution works for him.

Can you clarify the idea with `set(STM32_TOOLCHAIN_PATH "${CMAKE_CURRENT_SOURCE_DIR}/tools/gcc-arm-none-eabi-latest" CACHE)` ? Wouldn't that hardcode the toolchain location to a specific path? The goal of this PR is that the old way to find the toolchain which used `NO_DEFAULT_PATH` still works.

atsju:
`NO_DEFAULT_PATH` is the old behavior yes, but I think this is not the way to do. The "new" behavior must work easily for everybody. If it's not working we must first understand why before doing any correction. 
Here it seems the PATH variable is checked before cmake variable. But if you read the documentation CACHE variables are evaluated even before PATH so maybe it's just enough to `set(STM32_TOOLCHAIN_PATH "${CMAKE_CURRENT_SOURCE_DIR}/tools/gcc-arm-none-eabi-latest" CACHE)`.

I think your solution would work for Zisi but I don't think it's the best approach (maybe I'm wrong).

robamu:
Hmm. The solution above is then something inside the `CMakeLists.txt` of the user, right?

Another option: `set(STM32_TOOLCHAIN_PATH ${STM32_TOOLCHAIN_PATH} CACHE)` if `STM32_TOOLCHAIN_PATH` is set to it takes precedence

atsju:
> Hmm. The solution above is then something inside the `CMakeLists.txt` of the user, right?

right

> Another option: `set(STM32_TOOLCHAIN_PATH ${STM32_TOOLCHAIN_PATH} CACHE)` if `STM32_TOOLCHAIN_PATH` is set to it takes precedence

Reading [documentation](https://cmake.org/cmake/help/latest/command/find_program.html) for 5th time at least. I now think Zisi is in case where 5 takes precedance over 7. Maybe we should use `HINTS` instead `PATH` then 4 would take precedance over 5.

If you have some time maybe do a second branch using `HINTS` that Zisi could test.

robamu:
`HINTS` is the perfect solution I think. I'm testing it but it appears to do the correct thing: `STM32_TOOLCHAIN_PATH` takes precedence over anything in path

zisi:
I try 
```
find_program(CMAKE_C_COMPILER NAMES ${STM32_TARGET_TRIPLET}-gcc HINTS ${TOOLCHAIN_BIN_PATH})
find_program(CMAKE_CXX_COMPILER NAMES ${STM32_TARGET_TRIPLET}-g++ HINTS ${TOOLCHAIN_BIN_PATH})
find_program(CMAKE_ASM_COMPILER NAMES ${STM32_TARGET_TRIPLET}-gcc HINTS ${TOOLCHAIN_BIN_PATH})
```

and it works by setting the path of toolchain `set(STM32_TOOLCHAIN_PATH "${CMAKE_CURRENT_SOURCE_DIR}/tools/gcc-arm-none-eabi-latest")` and the system it has a toolchain in /usr/bin.

robamu:
I am preparing a new PR with hints

atsju:
Closed in favor of #266

PR #266, opened 2021-08-10
Using HINTS for STM Toolchain Path

### LINKED ISSUE #251: Could no set path of compiler

robamu:
This causes toolchains found via the STM32_TOOLCHAIN_PATH variable
to take precedence over a toolchain in the system path.
Also updated README accordingly

Fixes #251  . Closes #265


atsju (approved):

Hish15 (commented):
Hi, 

What is the confidence on the TOOLCHAIN_BIN_PATH correctness?

If I quote [cmake's find packages doc](https://cmake.org/cmake/help/v3.0/command/find_package.html) :
```
Search paths specified by the HINTS option. These should be paths computed by system introspection, such as a hint provided by the location of another item already found. Hard-coded guesses should be specified with the PATHS option.
```

atsju:
> Hi,
> 
> What is the confidence on the TOOLCHAIN_BIN_PATH correctness?
> 
> If I quote [cmake's find packages doc](https://cmake.org/cmake/help/v3.0/command/find_package.html) :
> 
> ```
> Search paths specified by the HINTS option. These should be paths computed by system introspection, such as a hint provided by the location of another item already found. Hard-coded guesses should be specified with the PATHS option.
> ```

Hi,
I think you want documentation of [find_program](https://cmake.org/cmake/help/latest/command/find_program.html) and not find_package. I agree spirit is exactly same. You can judge the confidence here :https://github.com/ObKo/stm32-cmake/blob/84155601b72ec3c278009d60376672dd3c8d4ad8/cmake/stm32/common.cmake#L16-L45

Maybe `PATH` is better than `HINTS` in the spirit but I doesn't work. See related issue.

Hish15:
Well, we want things to work so lets go for the HINT thing then. cmake's documentation is quite poor on the difference between PATH and HINTS anyway.

Issue #274, opened 2021-11-07
Blinky example won't start when changing blinky.c -> blinky.cpp


Osteri:
**Describe the bug**
C++ version of main will compile fine, but LED won't start flashing.

**Sources to reproduce**
I think it's so simple I won't provide a fork, but I can if it's really needed.

**Steps to reproduce**
1. Go to Blinky example project directory
2. Rename blinky.c -> blinky.cpp
2. Change the source file in CMakeLists.txt
3. Add CXX as the project language in CMakeLists.txt
4. Compile and flash your microcontroller (I'm using USB DFU if that makes a difference)
5. LED is not flashing as it should

**Expected behavior**
LED should flash as in the C example. Nothing will happen with C++ version. Something wrong with startup procedure on C++?

**Environment (please complete the following information):**
 - OS: Fedora 34
 - Compiler: arm-none-eabi-gcc (Fedora 10.2.0-4.fc34) 10.2.0
 - stm32-cmake: commit 12240e0
 - cmake: 3.19.7
 - HAL/cube/CMSIS: 3d7a3c1 Release v2.6.6, 313c279 Release v1.7.12, 96d6da4 Official ARM version: v5.6.0

Osteri:
If I'll add `STM32::Nano` target for linker (as guided in the docs), it will give me linker errors:
```
[ 87%] Linking CXX executable executable.elf
/usr/lib/gcc/arm-none-eabi/10.2.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/10.2.0/../../../../arm-none-eabi/lib/thumb/v7e-m+fp/hard/libc_nano.a(lib_a-abort.o): in function `abort':
/builddir/build/BUILD/newlib-4.1.0/build-nano/arm-none-eabi/thumb/v7e-m+fp/hard/newlib/libc/stdlib/../../../../../../../../newlib/libc/stdlib/abort.c:59: undefined reference to `_exit'
/usr/lib/gcc/arm-none-eabi/10.2.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/10.2.0/../../../../arm-none-eabi/lib/thumb/v7e-m+fp/hard/libc_nano.a(lib_a-signalr.o): in function `_kill_r':
[...]

PR #275, opened 2021-11-09
Duplicate blinky C version as C++ example

Osteri:
As discussed with @Hish15 and @atsju in #274, adding an C++ example would be nice for newcomers. 

So here it is.

Only difference compared to `blinky.c`:
```cpp
extern "C" {
  void SysTick_Handler();
}
```
Tested with F4.

### DIFF
--- a/examples/blinky-cpp/CMakeLists.txt
+++ b/examples/blinky-cpp/CMakeLists.txt
@@ -1,0 +1,82 @@
+cmake_minimum_required(VERSION 3.16)
+set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
+
+project(stm32-blinky CXX C ASM)
+set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
+
+# Configure here which STM32 target(s) to build
+option(BLINKY_F4_EXAMPLE "Compile F4 example" ON)
+option(BLINKY_F1_EXAMPLE "Compile F1 example" OFF)
+option(BLINKY_L0_EXAMPLE "Compile L0 example" OFF)
+
+set(HAL_COMP_LIST RCC GPIO CORTEX)
+set(CMSIS_COMP_LIST "")
+
+if(BLINKY_F4_EXAMPLE)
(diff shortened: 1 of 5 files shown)


Hish15:
This does duplicate a lot of things...

I think it's better to just write a blinky.cpp file, then give the option to use or not C++ on the CMakeLists file.

Another approach would be to create a file `system_config.c`, move the `SysTick_Handler` definition there, and then state in the CmakeLists, or blinky file, that it's possible to make blink.c ->blinky.cpp  and get thing running flawlessly.

A third approch is not to have this example depend on interrupt, no need to redefine `SysTick_Handler` once again.

What do you guys think? (@Osteri and @atsju)

atsju:
I'm against 3rd option as it would remove some knowledge from the already simple example.

Having blinky.c compiles in C++ or C depending on Cmake or on extension seems good.
You could do a blinky.h file to have the `extern C` on declaration in .h and not on definition in .c/.cpp but this is matter of taste.

Osteri:
To clarify: we would like to have separate blinky.cpp and blinky.c in blinky example folder and then based on some switches interselect between them during the compilation phase?

> then give the option to use or not C++ on the CMakeLists file.

What kind of option? Best option IMO is the project definiton `project(stm32-blinky CXX C ASM)`. If we derive this option somehow and customize based on that we can reflect "is this a C++ or C pure project?".

> Having blinky.c compiles in C++ or C depending on Cmake or on extension seems good.

I think CMake uses file extension to define how to compile the current file (it selects C or C++ compiler depending on the file extension). So it makes more sense IMO that we have different `blinky.cpp` for C++ and `blinky.c` for C compared to `blinky.c` which we would either compile to C++ or C.

Hish15:
When I say option, I mean cmake option as seen here https://github.com/ObKo/stm32-cmake/blob/2dc0456bc5317a467ea25fc4d36dbc8e2178bdee/examples/blinky/CMakeLists.txt#L8

Having blinky.c and blinky.cpp and adding one or the over using target_sources(), and the previously defined option is the way to go yes. This option can be used to define le project conditionally too.

atsju:
close in favor of #285

Issue #248, opened 2021-07-27
findBSP is not run in actions


atsju:
There is a file tests/bsp/CMakeLists.txt but it's not executed by actions. 
It needs review and update to have all families before running.

PR #283, opened 2022-02-04
MP1 support

Inujel:
This is still very much a work in progress. 

I create the PR because I'd like to discuss a few points before going further 🙂. 

Could you please give me some feedback about the comments below? 
Thank you very much!


Inujel:
As observed [here](https://github.com/ObKo/stm32-cmake/pull/186#issuecomment-834422794), there are no repositories for the CMSIS and HAL drivers for the MP1.

Here's what I did in `utilities.cmake`:
```cmake
# A CMSIS or HAL driver can specify 'cube' as version number to indicate that the driver is taken from the Cube repository
set(STM32_FETCH_FAMILIES       F0      F1      F2      F3      F4       F7      G0      G4      H7       L0      L1      L4      L5     MP1    U5     WB      WL    )
set(STM32_FETCH_CUBE_VERSIONS  v1.11.2 v1.8.4  v1.9.3  v1.11.2 v1.26.1  v1.16.1 v1.4.1  v1.4.0  v1.9.0   v1.12.0 v1.10.3 v1.17.0 v1.4.0 1.5.0  v1.0.0 v1.12.0 v1.1.0)
set(STM32_FETCH_CMSIS_VERSIONS v2.3.5  v4.3.3  v2.2.5  v2.3.5  v2.6.6   v1.2.6  v1.4.0  v1.2.1  v1.10.0  v1.9.1  v2.3.2  v1.7.1  v1.0.4 cube   v1.0.0 v1.9.0  v1.1.0)
set(STM32_FETCH_HAL_VERSIONS   v1.7.5  v1.1.8  v1.2.7  v1.5.5  v1.7.12  v1.2.9  v1.4.1  v1.2.1  v1.10.0  v1.10.4 v1.4.4  v1.13.0 v1.0.4 cube   v1.0.0 v1.9.0  v1.1.0)
```
Is that ok?

Inujel:
I changed 
```
FetchContent_GetProperties(${CMSIS_NAME_L} POPULATED CMSIS_POPULATED)
if(NOT CMSIS_POPULATED)
		set(FETCHCONTENT_QUIET FALSE) # To see progress
		FetchContent_Populate(${CMSIS_NAME})
endif()
```
and similar to
```
FetchContent_MakeAvailable(${CMSIS_NAME})
```

Because
1. It's advised by the [CMake doc](https://cmake.org/cmake/help/latest/module/FetchContent.html)
2. There was a bug when `stm32_fetch_cube` was called multiple times from different scopes. After the first call:
	- `STM32_CUBE_${FAMILY}_PATH` is not set because we're in another scope (so the `continue()` path is not taken)
	- `FetchContent_Populate` is not called a second time, so `${CUBE_NAME_L}_SOURCE_DIR` is not set
	- `STM32_CUBE_${FAMILY}_PATH` is set to an empty string in the parent scope
	Using `FetchContent_MakeAvailable` always sets `${CUBE_NAME_L}_SOURCE_DIR`

Inujel:
Regarding this:
```md
CMSIS creates the following targets:

* `CMSIS::STM32::<FAMILY>` (e.g. `CMSIS::STM32::F4`) - common includes, compiler flags and defines for family
* `CMSIS::STM32::<TYPE>` (e.g. `CMSIS::STM32::F407xx`) - common startup source for device type, depends on `CMSIS::STM32::<FAMILY>`
* `CMSIS::STM32::<DEVICE>` (e.g. `CMSIS::STM32::F407VG`) - linker script for device, depends on `CMSIS::STM32::<TYPE>`

So, if you don't need linker script, you can link only `CMSIS::STM32::<TYPE>` library and provide your own script using `stm32_add_linker_script` function
```

### About the startup file
[CMSIS provides stm32mp15?axx.s and stm32mp15?cxx.s files](https://github.com/STMicroelectronics/STM32CubeMP1/tree/1.5.0/Drivers/CMSIS/Device/ST/STM32MP1xx/Source/Templates/gcc) but no stm32mp15?dxx.s nor stm32mp15?fxx.s (I have no clue why). 
It also provides a stm32mp15xx.s file which seems to be compatible for all MP15 devices (could you give me your opinion on this?). 

I think we can simply use stm32mp15xx.s for all MP15 devices, making `CMSIS::STM32::<TYPE>` targets redundant


### About the linker script
I think that the linker script should be managed differently for the MP1 devices because instead of having a FLASH and a RAM, they have a retention RAM, and 4 RAM sections that are configurable : [STM32MP15 RAM mapping](https://wiki.st.com/stm32mpu/wiki/STM32MP15_RAM_mapping).

This is important because the developer must ensure that the RAM layout used by the linker script matches the device tree used on the A7 side.

All MP15 have the same amount of RAM.

I think that
 - There's only sense in configuring the linker per target. There's no use having `CMSIS::STM32::<DEVICE>` targets.
 - We should provide a linker script generator that takes the sizes of the RAM sections in input and support 3 use cases :
    1. Default RAM geometry:
        ```cmake
        add_executable(blinky main.cpp)
        target_link_libraries(blinky PRIVATE CMSIS::STM32::MP1::M4 CMSIS::STM32::MP1::M4::DefaultLinkerScript)
        ```
    2. Custom RAM geometry with generated linker script:
        ```cmake
        add_executable(blinky main.cpp)
        stm32_make_mp1_linker_script(BLINKY_LINKER_SCRIPT ...<ram geometry description>)
        target_link_libraries(blinky PRIVATE CMSIS::STM32::MP1 BLINKY_LINKER_SCRIPT)
        ```
    3. Custom linker script:
        ```cmake
        add_executable(blinky main.cpp)
        target_link_libraries(blinky PRIVATE CMSIS::STM32::MP1)
        stm32_add_linker_script(blinky PRIVATE my_linker_script.ld)
        ```
 
The linker script for MP1 devices is quite different from what's produced by `linker_ld.cmake`, so I think it should be easier to write another generator.

### Conclusion about targets
I don't see any point in having targets other than `CMSIS::STM32::MP1::CM4`, they will all do the exact same thing in the end.

cmake/stm32/utilities.cmake
```diff
@@ -44,10 +44,11 @@ endfunction()
...
-set(STM32_FETCH_CMSIS_VERSIONS v2.3.5  v4.3.3  v2.2.5  v2.3.5  v2.6.6   v1.2.6  v1.4.0  v1.2.1  v1.10.0  v1.9.1  v2.3.2  v1.7.1  v1.0.4 v1.0.0 v1.9.0  v1.1.0)
-set(STM32_FETCH_HAL_VERSIONS   v1.7.5  v1.1.8  v1.2.7  v1.5.5  v1.7.12  v1.2.9  v1.4.1  v1.2.1  v1.10.0  v1.10.4 v1.4.4  v1.13.0 v1.0.4 v1.0.0 v1.9.0  v1.1.0)
+# A CMSIS or HAL driver can specify 'cube' as version number to indicate that the driver is taken from the Cube repository
+set(STM32_FETCH_FAMILIES       F0      F1      F2      F3      F4       F7      G0      G4      H7       L0      L1      L4      L5     MP1    U5     WB      WL    )
+set(STM32_FETCH_CUBE_VERSIONS  v1.11.2 v1.8.4  v1.9.3  v1.11.2 v1.26.1  v1.16.1 v1.4.1  v1.4.0  v1.9.0   v1.12.0 v1.10.3 v1.17.0 v1.4.0 1.5.0  v1.0.0 v1.12.0 v1.1.0)
+set(STM32_FETCH_CMSIS_VERSIONS v2.3.5  v4.3.3  v2.2.5  v2.3.5  v2.6.6   v1.2.6  v1.4.0  v1.2.1  v1.10.0  v1.9.1  v2.3.2  v1.7.1  v1.0.4 cube   v1.0.0 v1.9.0  v1.1.0)
+set(STM32_FETCH_HAL_VERSIONS   v1.7.5  v1.1.8  v1.2.7  v1.5.5  v1.7.12  v1.2.9  v1.4.1  v1.2.1  v1.10.0  v1.10.4 v1.4.4  v1.13.0 v1.0.4 cube   v1.0.0 v1.9.0  v1.1.0)
```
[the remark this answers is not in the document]
atsju: Pleas note : I did delete your commentsand create comments so that I can answer to each post with history

As for the modification, I did not look into detail but yes seems fine.

cmake/stm32/utilities.cmake
```diff
@@ -94,62 +106,55 @@ function(stm32_fetch_cube)
             continue()
         endif()
         
-		FetchContent_GetProperties(${CUBE_NAME} POPULATED CUBE_POPULATED)
-        if(NOT CUBE_POPULATED)
-            set(FETCHCONTENT_QUIET FALSE) # To see progress
-            FetchContent_Populate(${CUBE_NAME})
-        endif()
-        
+		FetchContent_MakeAvailable(${CUBE_NAME})
```
atsju, line 109: Original comment from @Inujel 
I changed 
```
FetchContent_GetProperties(${CMSIS_NAME_L} POPULATED CMSIS_POPULATED)
if(NOT CMSIS_POPULATED)
		set(FETCHCONTENT_QUIET FALSE) # To see progress
		FetchContent_Populate(${CMSIS_NAME})
endif()
```
and similar to
```
FetchContent_MakeAvailable(${CMSIS_NAME})
```

Because
1. It's advised by the [CMake doc](https://cmake.org/cmake/help/latest/module/FetchContent.html)
2. There was a bug when `stm32_fetch_cube` was called multiple times from different scopes. After the first call:
	- `STM32_CUBE_${FAMILY}_PATH` is not set because we're in another scope (so the `continue()` path is not taken)
	- `FetchContent_Populate` is not called a second time, so `${CUBE_NAME_L}_SOURCE_DIR` is not set
	- `STM32_CUBE_${FAMILY}_PATH` is set to an empty string in the parent scope

Using `FetchContent_MakeAvailable` always sets `${CUBE_NAME_L}_SOURCE_DIR`

atsju: You seem to have worked on the subject. I do not see any drawback, just ensure the example are still building.
It's good work 👍

cmake/FindCMSIS.cmake
```diff
@@ -218,9 +223,17 @@ foreach(COMP ${CMSIS_FIND_COMPONENTS_FAMILIES})
...
         string(TOLOWER ${DEVICE} DEVICE_L)
         string(TOLOWER ${TYPE} TYPE_L)
+
+        set(STARTUP_NAMES startup_stm32${TYPE_L}.s startup_stm32${TYPE_L}${CORE_Ucm}.s)
+        if(${FAMILY} STREQUAL "MP1")
+            # There are stm32mp15?axx.s and stm32mp15?cxx.s files but no stm32mp15?dxx.s nor stm32mp15?fxx.s. No idea why.
+            # I think that stm32mp15xx.s should be compatible with all stm32mp15 devices anyway.
+            # This might need refinement if devices other stm32mp15 are released.
+            list(APPEND STARTUP_NAMES startup_stm32mp15xx.s)
+        endif()
```
atsju, line 233: Original comment from @Inujel 
Regarding this:
```md
CMSIS creates the following targets:

* `CMSIS::STM32::<FAMILY>` (e.g. `CMSIS::STM32::F4`) - common includes, compiler flags and defines for family
* `CMSIS::STM32::<TYPE>` (e.g. `CMSIS::STM32::F407xx`) - common startup source for device type, depends on `CMSIS::STM32::<FAMILY>`
* `CMSIS::STM32::<DEVICE>` (e.g. `CMSIS::STM32::F407VG`) - linker script for device, depends on `CMSIS::STM32::<TYPE>`

So, if you don't need linker script, you can link only `CMSIS::STM32::<TYPE>` library and provide your own script using `stm32_add_linker_script` function
```

### About the startup file
[CMSIS provides stm32mp15?axx.s and stm32mp15?cxx.s files](https://github.com/STMicroelectronics/STM32CubeMP1/tree/1.5.0/Drivers/CMSIS/Device/ST/STM32MP1xx/Source/Templates/gcc) but no stm32mp15?dxx.s nor stm32mp15?fxx.s (I have no clue why). 
It also provides a stm32mp15xx.s file which seems to be compatible for all MP15 devices (could you give me your opinion on this?). 

I think we can simply use stm32mp15xx.s for all MP15 devices, making `CMSIS::STM32::<TYPE>` targets redundant


### About the linker script
I think that the linker script should be managed differently for the MP1 devices because instead of having a FLASH and a RAM, they have a retention RAM, and 4 RAM sections that are configurable : [STM32MP15 RAM mapping](https://wiki.st.com/stm32mpu/wiki/STM32MP15_RAM_mapping).

This is important because the developer must ensure that the RAM layout used by the linker script matches the device tree used on the A7 side.

All MP15 have the same amount of RAM.

I think that
 - There's only sense in configuring the linker per target. There's no use having `CMSIS::STM32::<DEVICE>` targets.
 - We should provide a linker script generator that takes the sizes of the RAM sections in input and support 3 use cases :
    1. Default RAM geometry:
        ```cmake
        add_executable(blinky main.cpp)
        target_link_libraries(blinky PRIVATE CMSIS::STM32::MP1::M4 CMSIS::STM32::MP1::M4::DefaultLinkerScript)
        ```
    2. Custom RAM geometry with generated linker script:
        ```cmake
        add_executable(blinky main.cpp)
        stm32_make_mp1_linker_script(BLINKY_LINKER_SCRIPT ...<ram geometry description>)
        target_link_libraries(blinky PRIVATE CMSIS::STM32::MP1 BLINKY_LINKER_SCRIPT)
        ```
    3. Custom linker script:
        ```cmake
        add_executable(blinky main.cpp)
        target_link_libraries(blinky PRIVATE CMSIS::STM32::MP1)
        stm32_add_linker_script(blinky PRIVATE my_linker_script.ld)
        ```
 
The linker script for MP1 devices is quite different from what's produced by `linker_ld.cmake`, so I think it should be easier to write another generator.

### Conclusion about targets
I don't see any point in having targets other than `CMSIS::STM32::MP1::CM4`, they will all do the exact same thing in the end.

Inujel: Alirght! Thanks for the review.

Inujel:
It looks like for the MP1, CMSIS is dependant on HAL : [system_stm32mp1xx.c includes stm32mp1xx_hal.h](https://github.com/STMicroelectronics/STM32CubeMP1/blob/817e844fa791a34cf67df8100345f2a7382373d7/Drivers/CMSIS/Device/ST/STM32MP1xx/Source/Templates/system_stm32mp1xx.c#L48)

CMSIS cannot be used without HAL for the MP1. I don't think that stm32-cmake should do anything special about it. 
The CMSIS test cannot succeed. I remove MP1 from the default list and add a warning about it.
```
if(NOT TEST_FAMILIES)
    set(TEST_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4 L5 U5 WB WL)
endif()

if("MP1" IN_LIST TEST_FAMILIES)
    message(WARNING "CMSIS for MP1 devices requires HAL, this test is expected to fail for the MP1 family")
endif()
```

Inujel:
Hi @atsju

It's ready to be reviewed.

cmake/stm32/common.cmake
```diff
@@ -205,6 +210,8 @@ function(stm32_get_memory_info)
         set(FLASH "32K")
     elseif(SIZE_CODE STREQUAL "8")
         set(FLASH "64K")
+    elseif(SIZE_CODE STREQUAL "A")
```
atsju, line 213: I think this is not completely correct. all size codes are incorrect for MP1 family. It must be managed explicitly and not just removing a warning.

cmake/FindCMSIS.cmake
```diff
@@ -72,21 +77,27 @@ function(cmsis_generate_default_linker_script FAMILY DEVICE CORE)
     stm32_get_memory_info(FAMILY ${FAMILY} DEVICE ${DEVICE} CORE ${CORE} HEAP SIZE HEAP_SIZE)
     stm32_get_memory_info(FAMILY ${FAMILY} DEVICE ${DEVICE} CORE ${CORE} STACK SIZE STACK_SIZE)
```
atsju, line 78: this should all be included in the `if MP1` as you won't use the information for creating the ld script

cmake/stm32/mp1.cmake
```diff
@@ -0,0 +1,26 @@
+set(STM32_MP1_TYPES 
+    MP151Axx MP151Cxx MP151Dxx MP151Fxx
+    MP153Axx MP153Cxx MP153Dxx MP153Fxx 
+    MP157Axx MP157Cxx MP157Dxx MP157Fxx)
+
```
atsju, line 5: look at F3 family. If I remember correctly this must match the existing startup scripts

cmake/stm32/utilities.cmake
```diff
@@ -90,66 +102,59 @@ function(stm32_fetch_cube)
...
     foreach(FAMILY ${ARGV})
-        set(HAL_NAME STM32-HAL-${FAMILY})
-        string(TOLOWER ${HAL_NAME} HAL_NAME_L)
-        
-        if(STM32_HAL_${FAMILY}_PATH)
-            message(INFO "STM32_HAL_${FAMILY}_PATH specified, skipping fetch for ${HAL_NAME}")
-            continue()
+        if(STM32_USE_HAL_FROM_CUBE_${FAMILY})
+            stm32_fetch_cube(${FAMILY})
+            message("Cube fetched for ${FAMILY} at ${STM32_CUBE_${FAMILY}_PATH}")
```
atsju, line 145: Please also add the message for CMSIS to stay coherent
```suggestion
            message(STATUS "Cube fetched for ${FAMILY} at ${STM32_CUBE_${FAMILY}_PATH}")
```

tests/fetch/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
 
 if(NOT TEST_FAMILIES)
-    set(TEST_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4 L5 U5 WB WL)
+    set(TEST_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4 L5 MP1 U5 WB WL)
```
atsju, line 5: hal and bsp are not tested for MP1 family ?

tests/cmsis/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ if(NOT TEST_FAMILIES)
     set(TEST_FAMILIES F0 F1 F2 F3 F4 F7 G0 G4 H7 L0 L1 L4 L5 U5 WB WL)
 endif()
 
+if("MP1" IN_LIST TEST_FAMILIES)
+    message(WARNING "CMSIS for MP1 devices requires HAL, this test is expected to fail for the MP1 family")
+endif()
```
atsju, line 10: Can you please share what is failing ? Just for my understanding

cmake/FindBSP.cmake
```diff
@@ -288,7 +288,7 @@ foreach(COMP ${BSP_FIND_COMPONENTS})
     string(TOLOWER ${COMP} COMP_L)
     string(TOUPPER ${COMP} COMP_U)
     
-    string(REGEX MATCH "^STM32([FGHLUW][0-9BL])([0-9A-Z][0-9M][A-Z][0-9A-Z])?_?(M0PLUS|M4|M7)?.*$" COMP_U ${COMP_U})
+    string(REGEX MATCH "^STM32([FGHLMUW]P?[0-9BL])([0-9A-Z][0-9M][A-Z][0-9A-Z])?_?(M0PLUS|M4|M7)?.*$" COMP_U ${COMP_U})
```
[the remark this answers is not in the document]
Inujel: Done

cmake/FindCMSIS.cmake
```diff
@@ -218,9 +229,17 @@ foreach(COMP ${CMSIS_FIND_COMPONENTS_FAMILIES})
         stm32_get_chip_type(${FAMILY} ${DEVICE} TYPE)
         string(TOLOWER ${DEVICE} DEVICE_L)
         string(TOLOWER ${TYPE} TYPE_L)
+
+        set(STARTUP_NAMES startup_stm32${TYPE_L}.s startup_stm32${TYPE_L}${CORE_Ucm}.s)
+        if(${FAMILY} STREQUAL "MP1")
+            # There are stm32mp15?axx.s and stm32mp15?cxx.s files but no stm32mp15?dxx.s nor stm32mp15?fxx.s. No idea why.
+            # I think that stm32mp15xx.s should be compatible with all stm32mp15 devices anyway.
+            # This might need refinement if devices other stm32mp15 are released.
+            list(APPEND STARTUP_NAMES startup_stm32mp15xx.s)
```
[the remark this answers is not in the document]
Inujel: Sorry about that. I remember changing it. I guess I screwed up something when committing 😅 .
This is fixed.

I admit that I did not understand everything about types and families.
I will look into `stm32_get_chip_type`, but can you tell me what feels wrong to you?

cmake/FindCMSIS.cmake
```diff
@@ -72,21 +77,27 @@ function(cmsis_generate_default_linker_script FAMILY DEVICE CORE)
...
-            -DRAM_SHARE_SIZE="${RAM_SHARE_SIZE}" 
-            -DSTACK_SIZE="${STACK_SIZE}" 
-            -DHEAP_SIZE="${HEAP_SIZE}" 
-            -DLINKER_SCRIPT="${OUTPUT_LD_FILE}"
-            -P "${STM32_CMAKE_DIR}/stm32/linker_ld.cmake"
-    )
+    if(${FAMILY} STREQUAL MP1)
+        add_custom_command(OUTPUT "${OUTPUT_LD_FILE}"
+            COMMAND ${CMAKE_COMMAND}
+                -E copy ${STM32_CMAKE_DIR}/stm32/mp15xx.ld ${OUTPUT_LD_FILE})
```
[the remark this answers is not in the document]
Inujel: Ok, done!

cmake/stm32/common.cmake
Inujel: There's no flash on the MP1 family. The only size code is "A" and I though that's what it meant. Don't you agree?
I did not put that there just to remove a warning. 

To me, this part just don't apply for the MP1 family.

cmake/stm32/mp1.cmake
Inujel: That would mean that a STM32F157F chip would be of STM32C157 type? Is this really what we want?

tests/cmsis/CMakeLists.txt
Inujel: This is because `system_stm32mp1xx.c` (in CMSIS) includes `stm32mp1xx_hal.h` without a `#ifdef USE_HAL_DRIVER` guard.

cmake/stm32/utilities.cmake
Inujel: Ok

cmake/stm32/mp1.cmake
atsju: from my understanding yes this is what we want. The "type" corresponds to linker script type. Type reflects the peripherals only. While same chip exists with different size and package it doesn't impact the type

tests/cmsis/CMakeLists.txt
atsju: thanks.
I opened https://github.com/STMicroelectronics/STM32CubeMP1/pull/12 to ST as this is a n issue on their side from my point of view. So bad they have no "issue" tab for MP1 series. In the meantime I agree to let the warning

cmake/stm32/common.cmake
atsju: in STM32G473RE the E is the size code. 
Please also have a look at datasheet. From my point of view, MP1 has no such thing as a size_code.
maybe you should do same as L1 family (see `stm32l1_get_memory_info`) to return always 0. 
All these information makes me think a modification is needed.
image

cmake/FindCMSIS.cmake
Inujel: Done

tests/fetch/CMakeLists.txt
Inujel: Fixed

cmake/FindCMSIS.cmake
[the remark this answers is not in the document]
atsju: as discussed previously the correct startup could be used instead the generic one.

cmake/FindCMSIS.cmake
```diff
@@ -218,9 +235,17 @@ foreach(COMP ${CMSIS_FIND_COMPONENTS_FAMILIES})
         stm32_get_chip_type(${FAMILY} ${DEVICE} TYPE)
         string(TOLOWER ${DEVICE} DEVICE_L)
         string(TOLOWER ${TYPE} TYPE_L)
+
+        set(STARTUP_NAMES startup_stm32${TYPE_L}.s startup_stm32${TYPE_L}${CORE_Ucm}.s)
+        if(${FAMILY} STREQUAL "MP1")
+            # stm32mp15?dxx.s uses stm32mp15?axx.s
+            # stm32mp15?fxx.s uses stm32mp15?cxx.s
+            string(REPLACE "dxx" "axx" STARTUP_NAMES ${STARTUP_NAMES})
+            string(REPLACE "fxx" "cxx" STARTUP_NAMES ${STARTUP_NAMES})
```
atsju, line 244: this is not what I expected. If I'm correct it will work out of the box once you correct the "type" thing in `MP1.cmake`

atsju:
We are almost done. Seems very promising.

cmake/stm32/mp1.cmake
Inujel: Ok!

cmake/FindCMSIS.cmake
Inujel: Understood!

Inujel:
There's still something wrong in the fetch test but I can't reproduce it on my machine. I'm searching...

cmake/stm32/devices.cmake
```diff
@@ -1095,12 +1095,18 @@ set(STM32_ALL_DEVICES
...
+    MP151DA
+    MP151FA
+    MP153AA
+    MP153CA
+    MP153DA
+    MP153FA
+    MP157AA
+    MP157CA
+    MP157DA
+    MP157FA
```
atsju, line 1109: this is strange regarding datasheet but correct regarding the current STM32cmake implementation. This is also what made you use the A as size code... 
Is it possible to put MP151C instead MP151CA or will it break everything ?If it breaks, it's OK to leave as is.

cmake/stm32/mp1.cmake
```diff
@@ -0,0 +1,31 @@
+set(STM32_MP1_TYPES 
+    MP151Axx MP151Cxx
+    MP153Axx MP153Cxx 
+    MP157Axx MP157Cxx)
+
+set(STM32_MP1_TYPE_MATCH
+    "MP151[AD]A.?" "MP151[CF]A.?"
```
atsju, line 7: I have a slight preference for this :
```suggestion
    "MP151[AD]..?" "MP151[CF]..?"  
```
But regarding previous comment it might be better to leave as is 👍

atsju (approved):
For me this is good to merge. If you feel like BSP will be fixed later then remove it from the tests so that it doesn't fail .
I let you fix the fetch also.

@Hish15 we will need your review for this

After merge will open an issue for BSP test not working and CMSIS test that needs ST correction.

cmake/stm32/devices.cmake
Inujel: > If you feel like BSP will be fixed later then remove it from the tests so that it doesn't fail 
We probably need to remove the H7, WL and WB families as well then, they also fail in the same way.
The BSP test is not executed in the github actions.

> This is also what made you use the A as size code..
Yep :)

> Is it possible to put MP151C instead MP151CA or will it break everything ?
It will break because I used 
```cmake
set(STM32_MP1_TYPE_MATCH
     "MP151[AD]A.?" ...
```
But I don't mind at all changing to "MP151[AD](A.?)?", I agree that it'd be better.

atsju: > We probably need to remove the H7, WL and WB families as well then, they also fail in the same way.

I didn't see they fail silently. So we can let them this way (and there is already #248 to remember it. Add info there if necessary)

> But I don't mind at all changing to MP151[AD](A.?)?, I agree that it'd be better.

If we both agree it would be better please go ahead :relaxed:

Inujel: Done!

cmake/FindHAL.cmake
```diff
@@ -359,6 +378,7 @@ foreach(COMP ${HAL_FIND_COMPONENTS_FAMILIES})
                                                     STM32::${FAMILY}${CORE_C} 
                                                     CMSIS::STM32::${FAMILY}${CORE_C})
         target_include_directories(HAL::STM32::${FAMILY}${CORE_C} INTERFACE "${HAL_${FAMILY}${CORE_U}_INCLUDE}")
+        target_compile_definitions(HAL::STM32::${FAMILY}${CORE_C} INTERFACE USE_HAL_DRIVER)
```
Hish15, line 381: What is `USE_HAL_DRIVER` ?

atsju: Indeed, it's not needed if user includes `stm32${FAMILY_L}xx_hal.h`
It's only use is to include this file through file `Drivers\CMSIS\Device\ST\STM32MP1xx\Include\stm32${FAMILY_L}xx.h`

Inujel: Yes indeed, it seems to be a mistake. I remove it.

Hish15 (approved):
Good job!

atsju:
Yes well done! This was not an easy task and you were not the first trying out.
Also you asked good questions and had excellent proposals. Thank you.

Inujel:
Thanks for your kind words!

PR #285, opened 2022-02-19
Add blinky C++ example

### LINKED ISSUE #274: Blinky example won't start when changing blinky.c -> blinky.cpp

Hish15:
Replace PR close #275 and issue close #274.

### DIFF
--- a/examples/blinky/blinky.cpp
+++ b/examples/blinky/blinky.cpp
@@ -30,7 +30,9 @@ extern "C"
 
         // 1 Hz blinking
         if ((HAL_GetTick() % 500) == 0)
+        {
             HAL_GPIO_TogglePin(LED_PORT, LED_PIN);
+        }
     }
 }
 
--- a/examples/blinky/blinky.cpp
+++ b/examples/blinky/blinky.cpp
@@ -1,0 +1,62 @@
+#if defined STM32L0
+    #include <stm32l0xx_hal.h>
+
+    // STM32L0538-Discovery green led - PB4
+    #define LED_PORT                GPIOB
+    #define LED_PIN                 GPIO_PIN_4
+    #define LED_PORT_CLK_ENABLE     __HAL_RCC_GPIOB_CLK_ENABLE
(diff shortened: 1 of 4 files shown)


examples/blinky/blinky.cpp
```diff
@@ -0,0 +1,62 @@
...
+//This prevent name mangling for functions used in C/assembly files.
+extern "C"
+{
+    void SysTick_Handler(void)
+    {
+        HAL_IncTick();
+
+        // 1 Hz blinking
+        if ((HAL_GetTick() % 500) == 0)
+            HAL_GPIO_TogglePin(LED_PORT, LED_PIN);
```
atsju, line 33: same line or add {} but not this

Hish15: Not my code. Corrected it on the blinky.c file too.

atsju: I know ...

PR #287, opened 2022-02-22
Note in README about enabling the ASM language

Inujel:
If the user does not include the `ASM` language in his/her `project()` directive, the startup script is not compiled without warning and the error is not obvious.

Added a note about this in README.md.


Hish15:
Thanks for catching this @Inujel. This was a problem for startup file and linker script!
I did add a check on another PR #288. This way is safer. Closing this one.

Issue #193, opened 2021-04-22
Nucleo STM32F103RB being bricked, blinky example not running


Zataras:
Hi,
I'm trying to use your toolchain for STM32F103RB development, while I'm constantly having this issue.
I was trying with my own application first, but I decided to try with your blinky example.
I've removed F4 and L0 targets, and left only F1 with change to F103 device.
I'm configuring CMake with simple _cmake -DCMAKE_MODULE_PATH=../cmake ../blinky_
I'm generating hex file using extra target:
`ADD_CUSTOM_TARGET(${TARGET}.hex DEPENDS ${TARGET} COMMAND ${CMAKE_OBJCOPY} -Oihex ${FILENAME} ${FILENAME}.hex)`
After running _make blinky.hex_ I get successful build and the artifacts.
After flashing nucleo with generated binary, nucleo is no longer visible by flasher application. I need to switch to Windows and use ST's official app and flash with reset button pressed. This helps to nucleo being visible, but flashing hex file with official flasher also do not the app to start blink led.
Previosly I was using TrueStudio configuration and everything was working fine. 
Can you help?

atsju:
Hi @Zataras,
it will be very difficult to reproduce without more information (especially the stm32-cmake version). 
It would really help if you share a public git repo with the modifications you explained. I would like to mention that starting from stm32-cmake example as you did is very good because it minimizes "unknown" code to check and you already tested the "good flasher with bad hex" which is great also. 

I would recommend you fork this repo (if it's not already done), do the modifications you explained in text and then share the sha of the commit you want to debug. This way anybody will be able to build with exact same configuration.

Zataras:
Hi @atsju,
Thank you for response. That is good point, I will prepare public fork customized and share link here.

Zataras:
@atsju, I've prepared more details. I've built binary from commit _1a1e23ad095c4c0b1869fe12672a73483f55b597_ of [my fork](https://github.com/Zataras/stm32-cmake/commit/1a1e23ad095c4c0b1869fe12672a73483f55b597).
In default opt directory I got V1.8.3 / 20-October-2020 of STM32CubeF1. 
I'm using cmake version 3.20.1.
Some more info printed by CMake:
-- No STM32_TARGET_TRIPLET specified, using default: arm-none-eabi
-- The C compiler identification is GNU 10.2.1
-- The ASM compiler identification is GNU
-- Found assembler: /opt/gcc-arm/bin/arm-none-eabi-gcc

[...]

PR #289, opened 2022-05-07
Replaced __WFI with active infinite loop

Hish15:
In #193, it was discovered that __WFI can lead to unwanted sleep of the device.
This is the fix proposed by @atsju


atsju (approved):
you could also replace with `__NOP()` but this is fine

PR #291, opened 2022-05-31
set CMAKE_FIND_ROOT_PATH_MODE_PROGRAM

robamu:
This one should be [safe to set](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_PROGRAM.html). There are three other ones, but I'm not 100 % sure whether they can be hardcoded (although ONLY makes the most sense for me for the other ones):

1. `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY`
2. `CMAKE_FIND_ROOT_PATH_MODE_INCLUDE`
3. `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`

Setting the second one to ONLY is definitely problematic


Hish15:
Hello @robamu, what did you need to set this parameter ? What does it do ?

robamu:
It's an option generally recommended when cross-compiling. You can also see an example in the [CMake documentation](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#toolchain-files) . For cross-compiling, it generally only makes sense to search for programs in the host root filesystem. It's just set explicitely now, the default setting might be BOTH. Like already mentioned, the three other ones are more interesting for bare-metal cross compilation.

Hish15 (approved):

Issue #298, opened 2022-08-13
Check the cmake version and inform the user if it's not at least the requested one


Hish15:
This is linked to #297

PR #299, opened 2022-08-14
Check cmake version and warn user if lower than required

orangesalad:
Warn the user if the CMake version is less than 3.16. Fix for issue #298


Hish15 (approved):

cmake/stm32_gcc.cmake
```diff
@@ -1,3 +1,7 @@
+if(${CMAKE_VERSION} VERSION_LESS "3.16.0") 
+    message(WARNING "Current CMake verision is ${CMAKE_VERSION}. stm32-cmake requires CMake 3.16 or greater")
```
atsju, line 2: Funny, I do this each time also :)

```suggestion
    message(WARNING "Current CMake version is ${CMAKE_VERSION}. stm32-cmake requires CMake 3.16 or greater")

```

atsju (changes requested):

atsju (approved):

PR #302, opened 2022-10-12
HAL SCAN 2

### LINKED ISSUE #268: HAL drivers detection without knowing the list of expected drivers
**Is your feature request related to a problem? Please describe.**
as shown in #255 update of HAL could lead to new modules that are not automatically managed by the repo. 
Moreover adding them could cause issues with older versions of HAL repos.

**Describe the solution you'd like**
HAL files should be discovered automatically and not depending on HAL_DRIVERS_XX HAL_LL_DRIVERS_XX HAL_EX_DRIVERS_XX

Hish15:
This does a lot of changes, but I did clean commits.
I suggest reviewing commit by commit.

Previously some tests selected all families on find_package command by setting no family.
Then the test proceeded to downloading the HAL for only one family. This is no longer an option since the HAL files are needed to retrieve the component list for each family. I changed the tests to test feed only the tested family to the find_package command.

close #268

Tested on

- [x] STM32F4
- [x] STM32H7 
- [x] STM32L0


cmake/FindHAL.cmake
```diff
@@ -1,251 +1,89 @@
...
-    adc crc cryp dac dma flash hash i2c mmc opamp pcd pwr rcc
-    rng rtc sai sd smartcard smbus spi tim uart usart
-)
-set(HAL_LL_DRIVERS_U5
-    adc comp cordic crc crs dac dlyb dma dma2d exti fmac fmc gpio i2c icache lpgpio lptim lpuart opamp pka pwr rcc 
-    rng rtc sdmmc spi tim ucpd usart usb utils
-)
+# This function gets a list of hal_driver using a given prefix and suffix
+#
+# out_list_hal_drivers   list of hal_drivers foud
```
Hish15, line 5:
```suggestion
# out_list_hal_drivers   list of hal_drivers found
```

cmake/FindHAL.cmake
```diff
@@ -1,251 +1,89 @@
...
+    else()
+    endif()
 endforeach()
+message("P2H HAL drivers are ${HAL_DRIVERS}")
 list(REMOVE_DUPLICATES HAL_DRIVERS)
 list(REMOVE_DUPLICATES HAL_LL_DRIVERS)
 
-foreach(COMP ${HAL_FIND_COMPONENTS})
+#Checkinf all the requested drivers
+foreach(COMP ${HAL_FIND_COMPONENTS_UNHANDLED})
```
atsju, line 84: I really don't see what this is used for ?
if the family is unknown, we would enter here but what would it bring ?
And it's not in find_package that we select the different hal drivers (crc, gpio, ...) we only expect a family. I need more explanation for this

atsju (commented):
Thank you for extra comments.
Minor fixes needed but seems great

cmake/FindHAL.cmake
```diff
@@ -1,251 +1,89 @@
...
+    if(CMAKE_MATCH_1) #Matches the family part of the provided STM32<FAMILY>[..] component
+        get_list_hal_drivers(HAL_DRIVERS_${FAMILY} ${HAL_${FAMILY}_PATH} "hal")
+        get_list_hal_drivers(HAL_EX_DRIVERS_${FAMILY} ${HAL_${FAMILY}_PATH}  "ex")
+        get_list_hal_drivers(HAL_LL_DRIVERS_${FAMILY} ${HAL_${FAMILY}_PATH} "ll")
+        list(APPEND HAL_DRIVERS ${HAL_DRIVERS_${FAMILY}})
+        list(APPEND HAL_LL_DRIVERS ${HAL_LL_DRIVERS_${FAMILY}})
+    else()
+    endif()
 endforeach()
+message("P2H HAL drivers are ${HAL_DRIVERS}")
```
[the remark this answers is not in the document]
Hish15: This is not supposed to be left ^^

cmake/FindHAL.cmake
Hish15: Ok, I need to add more comments then

cmake/FindHAL.cmake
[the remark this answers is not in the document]
atsju: I mean we could leave more trace for debug. Here we have access to different debug levels and by default it will not be printed. I you needed it during development then it must be usefull at least at a `TRACE` level
I explained same to SBE for data-logging.

cmake/FindHAL.cmake
Hish15: I see your confusion. It's true that the Readme does not says that drivers can be specified in the find_package(HAL <>) call.
As the code is today it is possible to do it. Some examples do use this feature.
https://github.com/ObKo/stm32-cmake/blob/510328de664ae88e90fe9eb1697b9b0d8782c60d/examples/freertos/CMakeLists.txt#L49
https://github.com/ObKo/stm32-cmake/blob/510328de664ae88e90fe9eb1697b9b0d8782c60d/examples/blinky-ll/CMakeLists.txt#L8

I guess the only good option here is to update the Readme file. It does make sense to be able to specify the list of drivers.

atsju (approved):

PR #304, opened 2022-11-19
Increase ease of use with: autoset/find of `TOOLCHAIN_BIN_PATH` var

simoneruffini:
## Feature description
I wanted to increase the ease of use of this library for new beginners and to remove additional parameters that you would need to add either via command line, env variable or CMake. 

The subject in question is the "autofind" of the bin folder where the compiler is placed. 

I know that the user of the library could just define its own `STM32_TOOLCHAIN_PATH` but I'm asking why we can't have an already "precise location" and use that variable just for very special overrides like in case of multiple toolchains installed.

You could say that `/usr` is pretty default, but for non privileged users where installation start from $HOME or for nixos users (as myself)  that don't have the default LFS with all binaries in `/usr/bin`, this feature could be helpful and "hassle-free".

For now I just added a proof of work because I don't understand fully the use of the `STM32_TOOLCHAIN_PATH` variable. Hence I'm happy to hear suggestions on how to better implement this feature if considered useful.
There needs to be for certain a way to override this procedure (see later).

## Some observations
From what I understood this two variables:
```
set(TOOLCHAIN_INC_PATH "${STM32_TOOLCHAIN_PATH}/${STM32_TARGET_TRIPLET}/include")
set(TOOLCHAIN_LIB_PATH "${STM32_TOOLCHAIN_PATH}/${STM32_TARGET_TRIPLET}/lib")
```
Are not used inside the codebase, while for 
```
set(TOOLCHAIN_SYSROOT "${STM32_TOOLCHAIN_PATH}/${STM32_TARGET_TRIPLET}")
```
_seems_ to be debate (?) in this PR #290 , i.e.  by default that value points to a non existant path `/usr/arm-none-eabi`.
From what I understood the variable `STM32_TOOLCHAIN_PATH` has no advantage to the generation of  `TOOLCHAIN_BIN_PATH` and instead is usually used when code is compiled from source.  I don't think that `TOOLCHAIN_BIN_PATH` is strictly dependent of `STM32_TOOLCHAIN_PATH` but rather it could be when set (e.g. compile from source). For the majority of cases it could take the default path found in `$PATH` (hence my function), and when this is not the case leave the user access to override that value.

### DIFF
--- a/cmake/stm32/common.cmake
+++ b/cmake/stm32/common.cmake
@@ -14,6 +14,12 @@ foreach(FAMILY ${STM32_SUPPORTED_FAMILIES_LONG_NAME})
 endforeach()
 list(REMOVE_DUPLICATES STM32_SUPPORTED_FAMILIES_SHORT_NAME)
 
+
+if(NOT STM32_TARGET_TRIPLET)
+    set(STM32_TARGET_TRIPLET "arm-none-eabi")
+    message(STATUS "No STM32_TARGET_TRIPLET specified, using default: " ${STM32_TARGET_TRIPLET})
+endif()
+
 if(NOT STM32_TOOLCHAIN_PATH)
     if(DEFINED ENV{STM32_TOOLCHAIN_PATH})
         message(STATUS "Detected toolchain path STM32_TOOLCHAIN_PATH in environmental variables: ")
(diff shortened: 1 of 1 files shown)


cmake/stm32/common.cmake
```diff
@@ -33,16 +39,40 @@ if(NOT STM32_TOOLCHAIN_PATH)
...
-    set(STM32_TARGET_TRIPLET "arm-none-eabi")
-    message(STATUS "No STM32_TARGET_TRIPLET specified, using default: " ${STM32_TARGET_TRIPLET})
-endif()
+function(stm32_autoset_toolchain_bin_path STM32_TARGET_TRIPLET RESULT_VAL_TOOLCHAIN_BIN_PATH)
+# set the OS utility to find a program path
+    if(MINGW OR CYGWIN OR WIN32)
+        set(OS_UTIL_SEARCH_CMD where)
+    elseif(UNIX OR APPLE)
+        set(OS_UTIL_SEARCH_CMD which)
+    endif()
```
Hish15, line 48: Why don't you use cmake's  [find_program](https://cmake.org/cmake/help/v3.17/command/find_program.html) ?

Hish15 (commented):
Hello, Thank you for your contribution.
I will dive into the implications of your changes later.

cmake/stm32/common.cmake
simoneruffini: But can I use the result of find_program inside an execute_process block?

Moreover initially my idea was to use the shell for *nix systems. The proper way of finding a program is through the use of the builtin shell function `command -v` (on some barebone systems like certain busybox builds, `which` is not installed). However this would result in a non standard way of calling the search command for windows/Linux because you should call `sh -c command -v <here the program to find` and in windows you don't do that. So I opted to do it like this for now. If there is a better way we can change it.

Hish15:
Hi again, 

I had a look at your proposal and found that we are supposed to be doing what you propose.

The Readme.md file specifies that only `STM32_CUBE_<FAMILY>_PATH` is mandatory as a variable and defaults to `/opt/STM32Cube<FAMILY>`.
Then it's stated that STM32_TOOLCHAIN_PATH is also searched in the user's PATH:
https://github.com/ObKo/stm32-cmake/blob/6d07cb82bce578369984533a755908ea38b43b19/README.md?plain=1#L64

This is done by the calls to find_program in common.cmake and stm32_gcc.cmake. A hint is provided to those call. It will still look in the user's PATH. have a look at NO_DEFAULT_PATH here https://cmake.org/cmake/help/latest/command/find_program.html.

Do you find that to be not true/working in your case ?

Hish15:
I'm closing this then.

PR #309, opened 2023-01-07
stm32: Add support for some G0 devices

crvux:
Added some of G050xx, G051xx, G061xx and G0B0xx devices.


cmake/stm32/devices.cmake
```diff
@@ -560,18 +581,33 @@ set(STM32_ALL_DEVICES
...
+    G0B1VB
     G0B1VC
     G0B1VE
     G0C1CC
     G0C1CE
     G0C1KC
     G0C1KE
+    G0C1MC
+    G0C1ME
+    G0C1NE
```
atsju, line 610:
```suggestion
```
This one does not exist (today) https://www.st.com/en/microcontrollers-microprocessors/stm32g0x1.html

atsju (approved):
excellent job, thank you very much !

atsju:
after verification datasheet explicitly states the part exists... 
However some suppliers have stm32G0B1NE stock line (with 0 stock) but no stm32G0C1NE. This tends to confirm part doesn't exist. I will open an ST case tomorrow to ask unless you are used to it and can open it.

crvux:
For this patch I used exported product list xlsx from CubeMX. STM32G0B1NE truly have NA marketing status in this list, but I thought that there might be users who have this controller.

[g0.xlsx](https://github.com/ObKo/stm32-cmake/files/10368048/g0.xlsx)

atsju:
It's not very important anyway. This list is only used by a function wich prints list of supported devices. It's not very useful and a pain to maintain.
Everything will work fine as long as the other files you modified support the device.

PR #310, opened 2023-01-08
stm32: common: Fix 'stm32_generate_*_file'

crvux:
This patch takes into account that the .elf file name may differ from target name. Now target properties and generator expression used to determine the file path.


cmake/stm32/common.cmake
```diff
@@ -61,33 +61,44 @@ function(stm32_print_size_of_target TARGET)
 endfunction()
 
 function(stm32_generate_binary_file TARGET)
+    get_target_property(OUTPUT_NAME ${TARGET} OUTPUT_NAME)
+    get_target_property(BINARY_DIR ${TARGET} BINARY_DIR)
+    set(FILE_NAME "${OUTPUT_NAME}.bin")
+    set(FILE "${BINARY_DIR}/${FILE_NAME}")
```
atsju, line 67: Could you rename to have something more explicit like this (or if you have a better idea) ?:
```suggestion
    set(GEN_FILE_NAME "${OUTPUT_NAME}.bin")
    set(GEN_FILE "${BINARY_DIR}/${FILE_NAME}")
```

atsju (approved):
For me this is fine.
@Hish15  can you please review?

cmake/stm32/common.cmake
crvux: Yes, I will make it more explicit

crvux:
To comply with the DRY principle, I can move 'add_custom_command' to a more abstract function. Then, for example, 'stm32_generate_hex_file' might look like this
```cmake
function(stm32_generate_hex_file TARGET)
    stm32_generate_file(${TARGET} "hex" "ihex")
endfunction()
```
What do you think?

atsju:
> To comply with the DRY principle, I can move 'add_custom_command' to a more abstract function. Then, for example, 'stm32_generate_hex_file' might look like this
> 
> ```cmake
> function(stm32_generate_hex_file TARGET)
>     stm32_generate_file(${TARGET} "hex" "ihex")
> endfunction()
> ```
> 
> What do you think?

I will let @Hish15 answer for this one. Do not hesitate to make the proposition in a different PR. It sounds like a good idea

Hish15:
> To comply with the DRY principle, I can move 'add_custom_command' to a more abstract function. Then, for example, 'stm32_generate_hex_file' might look like this
> 
> ```cmake
> function(stm32_generate_hex_file TARGET)
>     stm32_generate_file(${TARGET} "hex" "ihex")
> endfunction()
> ```
> 
> What do you think?

Thanks for the PR @crvux 
I like this idea! Are you willing to do it this way ? algo can you add comments to the function? We are trying to improve the in code  documentation from 0 to the starts :D

crvux:
@Hish15 , @atsju can you please review new commits?

atsju (approved):
looks perfect. Really nice, well done 🥇

Hish15 (approved):

PR #315, opened 2023-03-05
Automatically enable used HAL driver modules

shoftmadder:
Currently, to add a driver module to a project, you have to:
- Add the driver module in `CMakeLists.txt`
- Set the driver module enabled (e.g. in a HAL config file, such as `stm32f4xx_hal_conf.h`)

This change automatically sets the driver module enable definition when a module is included in `CMakeLists.txt`, meaning that you don't also have to edit a header file to enable the module.

### DIFF
--- a/cmake/stm32/utilities.cmake
+++ b/cmake/stm32/utilities.cmake
@@ -44,9 +44,9 @@ include(FetchContent)
 
 # A CMSIS or HAL driver can specify 'cube' as version number to indicate that the driver is taken from the Cube repository
 set(STM32_FETCH_FAMILIES       F0      F1      F2      F3      F4       F7      G0      G4      H7       L0      L1      L4      L5     MP1    U5     WB      WL    )
-set(STM32_FETCH_CUBE_VERSIONS  v1.11.2 v1.8.4  v1.9.3  v1.11.2 v1.26.1  v1.16.1 v1.4.1  v1.4.0  v1.9.0   v1.12.0 v1.10.3 v1.17.0 v1.4.0 1.5.0  v1.0.0 v1.12.0 v1.1.0)
-set(STM32_FETCH_CMSIS_VERSIONS v2.3.5  v4.3.3  v2.2.5  v2.3.5  v2.6.6   v1.2.6  v1.4.0  v1.2.1  v1.10.0  v1.9.1  v2.3.2  v1.7.1  v1.0.4 cube   v1.0.0 v1.9.0  v1.1.0)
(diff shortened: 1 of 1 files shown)


Hish15:
Hello @shoftmadder, That's a nice catch ! I never noticed that!
(If you include the drivers files where you need them (#include "stm32f1xx_hal_dac.h" for instance), you don't see it.)

Can you change the exemples, in particular blinky, so it uses your change ?
I mean changing stm32xxxx_hal_conf.h files.

examples/blinky/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ if(BLINKY_F4_EXAMPLE)
     target_link_libraries(stm32-blinky-f4
         HAL::STM32::F4::RCC
         HAL::STM32::F4::GPIO
+        HAL::STM32::F4::FLASH
```
atsju, line 58: I believe there is a slight difference as now flash.c is compiled.
Before your change, only flash.h was included but flash.c not built.

atsju (approved):
> If it matters, I'm assigning all rights to the authors of this project

It doesn't matter for us. However, it makes me ask myself more questions than it answers :)

@hish15 please see my review comment and choose about merging

atsju:
@shoftmadder #317 workflow builds fine. You will have to find out why test fails on MP1 and G0 families with your change.

examples/blinky/CMakeLists.txt
atsju: It's not an issue from my point of view we can live with it. Will see with @Hish15 

However, have a look at workflow failure

atsju:
> Recommended solution -- move to a non-broken version of STM32CubeG0 in the test environment

It's not only for test environment. You must update lines 47-49 from utilities.cmake for doing that

About MP1: having `-D` by compiler and `#undef` in one file only, the result will depend on the how user includes files. 

**Thinking out loud:** use is asking for specific (sometimes all) components in `find_package(HAL`. in blinky it's `HAL_COMP_LIST`.
Maybe you can try to `target_compile_definitions` every requested component in each library of the family. For blinky this will lead to needing `FLASH` in `HAL_COMP_LIST` but not the actual library `HAL::STM32::F4::FLASH` with the sources. This would be more in the spirit of HAL. For users that do not specify specific HAL components but only the family it would behave as if all HAL were enabled (my typical lazy hal_conf.h just enables everything so this behaves the same). However sources would be built only if needed. 
I had a quick glance and it seems doable. Dos it seem reasonable ?

Hish15:
@shoftmadder Thanks for the changes, I will discuss this with @atsju. I think I understand your thought process,  but I'm not sure about the  `drop_excluded_hal_drivers` approach.
We will come back to you.

Cya

simoneruffini:
I would like to add that if you use a tool like STM32 CubeMX it automatically does that for you: changing the `stm32fxxx_hal_conf.h` file with the correctly enabled peripherals.

PR #317, opened 2023-03-11
add ability to manually trigger a workflow + run only on PR and master push

atsju:
this will also help find if #315 fails or master

### DIFF
--- a/examples/blinky/CMakeLists.txt
+++ b/examples/blinky/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.16)
 set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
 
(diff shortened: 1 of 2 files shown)


examples/blinky/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.16)
 set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/stm32_gcc.cmake)
 
+set(STM32_CUBE_F4_PATH "C:\\GIT\\STM32CubeF4")
```
Hish15, line 4: Use the `stm32_fetch_cube(F4)` function, in the if stm32_fetch_cube(F4) instead.

Hish15 (commented):
You pushed 2 commits intead of one here

examples/blinky/CMakeLists.txt
atsju: This was not supposed to be pushed. It slipped though on commit

Hish15:
@atsju, let's make the changes to merge this please

PR #331, opened 2023-11-14
Add STM32C0 support

xanthio:
This Add support for STM32C0 range MCU

PR #332, opened 2023-11-14
Add C0 MCU in supported devices families

xanthio:
Fix non updated README for C0 MCU in supported devices families in PR #331


atsju:
thank you

PR #336, opened 2024-04-08
Support STM32U0 Family

xanthio:
Bring STM32U0 family support to project


atsju (approved):
I did not check the Flash sizes as I usually do but this looks high quality PR. Thank you.
@Hish15 I let you merge if it's OK for you

xanthio:
> I did not check the Flash sizes as I usually do but this looks high quality PR. Thank you. @Hish15 I let you merge if it's OK for you

Actually, Flash size variants for this family is weird as we can find 16/32/64M of Flash but it depends the package:
    U031C6
    U031C8
    U031F4
    U031F6
    U031F8
    U031G6
    U031G8
    U031K4
    U031K8
    U031R6
    U031R8

Do you think there is an interest in splitting this subfamilly ?

atsju:
I checked your code is actually correct. 
example: U031C6 => U031 gives RAM size because is a specific subfamily and here all the subfamily has same RAM (cool, not always the case)  
C => package => no interrest for us in this case. Sometimes it has an effect  
6 => flash size. See file common.cmake

xanthio:
@Hish15 Is there an issue for not merging this PR ?

I'm starting a project with STM32U0 target and would really like to submodule this before.

Hish15:
Seems good to me, solve the conflict and we are good to go!

Hish15 (approved):

PR #337, opened 2024-04-16
Update to use latest Cube / HAL / CMSIS for F4

slide333333:
Latest released Cube version is v1.28.0: https://htmlpreview.github.io/?https://github.com/STMicroelectronics/STM32CubeF4/blob/master/Release_Notes.html
It uses HAL version 1.8.2 and CMSIS version 2.6.9


atsju:
Thank you for your contribution

PR #340, opened 2024-05-11
Huge refactoring to ease project maintenance

xanthio:
Here is a proposition that aims to ease project maintenability :
- Matrixing of the workflows is created dynamically so they will run for a new familly as soon as it is added
- Supported Devices and Family are now append in the <family>.cmake file to ease the addition of a new family
- ST Cube / CMSIS / HAL repos version are defined in the  <family>.cmake to avoid conflicts between 2 PR for 2 family versions updates.
- ST repos dependency are checked each night and a PR is created if a new version exists
- Every night a workflow runs to check for devices existing on ST website but not supported


xanthio:
Note: device STM32GBK1CB exists in supported devices list (g4.cmake) but I can't find where does it come from.

Hish15 (changes requested):
Hello, those changes are really welcome, thanks a lot!
That being said one single commit for so many changes is risky.
Can you split the changes into at leats 2/3 PR, one or two for the CI changes, and one for the changes to the families handling.
If something is broken too many changes at once is not ideal.

xanthio:
OK I will do so !

PR #342, opened 2024-05-24
Dynamic family matrix for CI jobs

xanthio:
In the objective of getting more contribution, these changes create the family matrix at CI run time.

I will be easier for a non CI-friendly contributor to add a new family to project as they won't have to make change in the workflow files.

Updating the Test workflow was the opportunity to update the deprecated checkout action step.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
       
     - name: Install ARM toolchain
       run: sudo apt-get install gcc-arm-none-eabi binutils-arm-none-eabi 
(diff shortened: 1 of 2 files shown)


xanthio:
I guess it would be a good idea to split Fetch/CMSIS/HAL in individual jobs too.

Hish15 (approved):

atsju:
Indeed, extensive comments are welcome. 
I would appreciate ELI5 comments but I let your @Hish15 decide if it's needed or not. As I understand, someone speaking Unix would not need any comment.

xanthio:
Actually its not so understandable for someone who is not a unix user.
I can split `Create matrix` step so the new steps names will comment the work that is done.

xanthio:
In fact not easy to split as steps use list variable that github action can not handle.
That is why I was using JSON object as recommended https://docs.github.com/en/actions/learn-github-actions/expressions#example-returning-a-json-object

PR #344, opened 2024-06-08
Fix multicore CMSIS tests

xanthio:
CMSIS tests was intended to create an executable for each core of each device of the family.
Unfortunately a typo in the CMakeFile only created an executable for main core.

Exemple:
WL family tests before:
image

WL family test after:
image

This PR gave the opportunity to get rid off not found warnings during CMSIS configuration step:
image


Hish15:
Good catch and fix, thx!

PR #347, opened 2024-06-27
Replace an had oc check for CMake's version with a builtin

LucasChollet:
`cmake_minimum_required` is the recommended way to do this check. Note that this will now make the configuration stop here rather than emiting a warning and panic later, on the line that actually require the version.

### DIFF
--- a/cmake/stm32_gcc.cmake
+++ b/cmake/stm32_gcc.cmake
@@ -1,7 +1,4 @@
-if(${CMAKE_VERSION} VERSION_LESS "3.16.0") 
-    message(WARNING "Current CMake version is ${CMAKE_VERSION}. stm32-cmake requires CMake 3.16 or greater")
-
-endif()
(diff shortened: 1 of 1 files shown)


Hish15:
The documentation state that this **has to be** set on the top cmakelists.txt file. Here we are on a toolchain file...
I'm not sure of the side effects of using `cmake_minimum_required` after the project call. This does more than checking the version, it changes the policies.

LucasChollet:
I was thinking that this was the same as when an external library call `cmake_minimum_required` after the main project call it for itself. Maybe this is different with a toolchain file... And it's not like we can ask kitware to standardize this eccentric workflow.
I don't have much to say except that it worked during my non-exhaustive testing. Feel free to close the PR if you feel like it.

Hish15:
I agree with you're reasoning, but lets not change things in a way discouraged by the documentation.

PR #348, opened 2024-09-24
add CMSIS DSP library target to FindCMSIS.cmake

zbrown23:
Add a target to the `FindCMSIS.cmake` that allows you to bring in the CMSIS DSP library for a given target. The targets end up looking like `CMSIS::STM32::H7::M7::DSP` as an example.

Let me know if there's anything you'd like me to tweak about this, otherwise it's a pretty simple change I think.

### DIFF
--- a/cmake/FindCMSIS.cmake
+++ b/cmake/FindCMSIS.cmake
@@ -313,6 +313,13 @@ foreach(COMP ${CMSIS_FIND_COMPONENTS_FAMILIES})
         set(CMSIS_${RTOS_COMP}_FOUND TRUE)
     endforeach()
 
+    if(NOT (TARGET CMSIS::STM32::${FAMILY}${CORE_C}::DSP}))
+        add_library(CMSIS::STM32::${FAMILY}${CORE_C}::DSP INTERFACE IMPORTED)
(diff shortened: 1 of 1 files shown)


cmake/FindCMSIS.cmake
```diff
@@ -313,6 +313,13 @@ foreach(COMP ${CMSIS_FIND_COMPONENTS_FAMILIES})
         set(CMSIS_${RTOS_COMP}_FOUND TRUE)
     endforeach()
 
+    if(NOT (TARGET CMSIS::STM32::${FAMILY}${CORE_C}::DSP}))
```
Hish15, line 316: At this point, do you know for sure that we do have the DSP for all families ? otherwise we might want to check first

zbrown23: Hi @Hish15, 
Sorry for taking so long to respond on this one. I went through and validated that every family has the ability to pull in DSP. Would you like me to write a unit test to make sure this remains the case?

Hish15: Of course @zbrown23, please add the UT, then we can merge this one year old PR.

PR #352, opened 2025-04-12
feat: allow user to define custom CMSIS version via CMake cache variable

ashen1queter:
Added support for setting a custom CMSIS version by defining the CMSIS_CUSTOM_VERSION cache variable. This allows users to override the default CMSIS version during CMake configuration by passing:

  -DCMSIS_CUSTOM_VERSION=<version>

If not specified, the script defaults to version v5.6.0

### DIFF
--- a/cmake/stm32/devices.cmake
+++ b/cmake/stm32/devices.cmake
@@ -160,13 +160,24 @@ macro(stm32_pretty_print_dev_list FAMILIES STM_DEVICES)
 endmacro()
 
 
-
 include(FetchContent)
 
+# Allow user to optionally define this via command line or script
+# Example: cmake -DCMSIS_CUSTOM_VERSION=5.9.0
+set(CMSIS_CUSTOM_VERSION "" CACHE STRING "Optional custom CMSIS version")
+
+# Use the custom version if provided, otherwise fall back to default
+if(DEFINED CMSIS_CUSTOM_VERSION AND NOT "${CMSIS_CUSTOM_VERSION}" STREQUAL "")
+    set(CMSIS_VERSION_TO_USE "v${CMSIS_CUSTOM_VERSION}")
+else()
+    set(CMSIS_VERSION_TO_USE "v5.6.0")
+endif()
+
+# FetchContent using the selected version
 FetchContent_Declare(
     STM32-CMSIS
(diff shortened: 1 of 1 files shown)


Hish15 (approved):

Hish15:
Hi @ashen1queter, It would be nice to add a UT before merging this. Are you willing to do it ? 
The idea is to check that this does indeed fetch a CMSIS version other than the default one when requested, and that it does retrieve the default too.

cmake/stm32/devices.cmake
```diff
@@ -160,13 +160,34 @@ macro(stm32_pretty_print_dev_list FAMILIES STM_DEVICES)
...
+# TODO: Update newer CMSIS versions when released 
+list(APPEND VALID_CMSIS_VERSIONS " 5.9.0" " 5.6.0_cm3" " 5.6.0_cm7" " 5.6.0_cm0" " 5.6.0_cm33" " 5.6.0" " 5.6.0_cm4" " 5.4.0" " 5.4.0_cm33" " 5.4.0_cm7")
+
+set(CMSIS_CUSTOM_VERSION "" CACHE STRING "Optional custom CMSIS version")
+
+# Use the custom version if provided, otherwise fall back to default
+if(DEFINED CMSIS_CUSTOM_VERSION AND NOT "${CMSIS_CUSTOM_VERSION}" STREQUAL "")
+    list(FIND VALID_CMSIS_VERSIONS " ${CMSIS_CUSTOM_VERSION}" VALID_VERSION_INDEX)
+    if(VALID_VERSION_INDEX EQUAL -1)
+        message(FATAL_ERROR "Invalid CMSIS version provided ${CMSIS_CUSTOM_VERSION}. Supported versions are: ${VALID_CMSIS_VERSIONS}")
```
Hish15, line 178: I like this, but it does make us follow the releases... Isn't letting the Fetch fail enough ?

What do you think @atsju ?

atsju: Same. Have an explicit error message on how the user can find valid versions when fetch fails (with one or two examples). 
But avoid having stm32-cmake to follow CMSIS releases. We have no automation for this.

ashen1queter: I will get back to this, I am currently involved in an internship

PR #355, opened 2025-05-04
Add all currently active MCUs from STM32C0 family

marcus-fre:
Verified by building and running a blinky on NUCLEO-C071RB


Hish15:
@atsju @ObKo,
For some reason I cannot rerun this PR, can one of you do it ?

PR #365, opened 2026-01-31
Fix LOAD Segment With RWX Permissions Warning

r2axz:
Starting with GCC11 LD shows the following error message when using the autogenerated linker script:

```
arm-none-eabi/bin/ld: warning: testi.elf has a LOAD segment with RWX permissions
```

This happens because some sections that go into the FLASH memory have read-write access permissions which in turn forces FLASH to become RWX.

Here is a relevant StackOverflow thread: https://stackoverflow.com/questions/73429929/gnu-linker-elf-has-a-load-segment-with-rwx-permissions-embedded-arm-project

ST suggest to change sections like

```
.preinit_array :
 {
   …
 } >FLASH
```

to

```
 .preinit_array (READONLY) :
 {
   …
 } >FLASH
```

Reference: https://wiki.st.com/stm32mcu/wiki/STM32CubeIDE:STM32CubeIDE_errata_1.15.x#General_issues
See 169316

This pull request adds the (READONLY) attribute to the corresponding sections if GCC version >= 11.


Hish15 (approved):
Nice!

atsju (approved):
Looks reasonable. Well documented, thank you !

